PR #47, opened 2017-07-07
filling the payload object

BinaryBeard:
When we init with a dictionary that was a JSON object, there is no Payload. We have to move all root level objects to the payload and convert them to data

# Test Case

Create a function in AWS Lambda

**My_Function**
```
'use strict';

exports.handler = (event, context, callback) => {
    console.log('Received event: %s', JSON.stringify(event, null, 2));
    vat object = event;
    event.another = "Test";
    event.object = { "hello": "world" };
    callback(null, object);
};
```

Using this AWS CLI command with the `input.txt`, you will get the response `outputfile.txt` below.

**AWS CLI lambda command**
```
aws lambda invoke \
--invocation-type RequestResponse \
--function-name My_Function \
--region us-east-1 \
--payload file://~/input.txt \
outputfile.txt
```

**input.txt**
```
{
    "action": "start"
}
```

**Command Line Output**
```
{
    "StatusCode": 200
}
```

**outputfile.txt**
```
{
    "action": "start",
    "another": "Test",
    "object": {
        "hello": "world"
    }
}
```

Using the Lambda portion of the SDK, `.payload = nil`. We expect:

```
.payload = [    
    "action": "start"
    "another": "Test"
    "object": [
        "hello": "world"
    ]
]
```

### DIFF
--- a/Sources/AWSSDKSwift/Services/lambda/Lambda_Shapes.swift
+++ b/Sources/AWSSDKSwift/Services/lambda/Lambda_Shapes.swift
@@ -873,10 +873,11 @@ extension Lambda {
         }
 
         public init(dictionary: [String: Any]) throws {
-            self.statusCode = dictionary["StatusCode"] as? Int32
-            self.logResult = dictionary["X-Amz-Log-Result"] as? String
-            self.payload = dictionary["Payload"] as? Data
-            self.functionError = dictionary["X-Amz-Function-Error"] as? String
+            var mutableDictionary = dictionary
+            self.statusCode = mutableDictionary.removeValue(forKey: "StatusCode") as? Int32
+            self.logResult = mutableDictionary.removeValue(forKey: "X-Amz-Log-Result") as? String
+            self.functionError = mutableDictionary.removeValue(forKey: "X-Amz-Function-Error") as? String
+            self.payload = mutableDictionary["Payload"] != nil ? mutableDictionary.removeValue(forKey: "Payload") as Data? : NSKeyedArchiver.archivedData(withRootObject: mutableDictionary)
         }
     }
 


noppoMan:
@BinaryBeard 
Thanks for your PR.
Actually almost files in AWSSDKSwift are automatically generated by [CodeGenerator](https://github.com/noppoMan/aws-sdk-swift/blob/master/Sources/CodeGenerator). 
(Although it is hard to find a bit, there is a description [DO NOT EDIT] on the top of the files.)
So if we want to change something, need to edit CodeGenerator.(Current code of it is too redundant to maintain. sorry.)

In order to solve this case, we may need to implement middleware or something like that to process the response body as well as request.

My an idea is here.

**Sources/AWSSDKSwift/Middlewares/lambda/LambdaResponseMiddleware.swift**
```swift
public protocol AWSResponseMiddleware {
    func chain(request: AWSRequest, request: Response) throws -> Response
}

struct LambdaResponseMiddleware: AWSResponseMiddleware {
    func chain(request: AWSRequest, request: Response) throws -> Response {
        switch request.operation {
        case "invoke":
             // code for processing response body
             // ..........
             return newResponse
         default:
             return response
        }
    }
}
```

**Sources/AWSSDKSwift/Services/lambda/Lambda_Shapes.swift**
```swift
public struct Lambda {

    let client: AWSClient

    public init(accessKeyId: String? = nil, secretAccessKey: String? = nil, region: AWSSDKSwiftCore.Region? = nil, endpoint: String? = nil) {
        self.client = AWSClient(
            accessKeyId: accessKeyId,
            secretAccessKey: secretAccessKey,
            region: region,
            service: "lambda",
            serviceProtocol: ServiceProtocol(type: .restjson),
            apiVersion: "2015-03-31",
            endpoint: endpoint,
            requestMiddlewares: [], // I changed here
            responseMiddlewares: [LambdaResponseMiddleware()], // I changed here
            possibleErrorTypes: [LambdaError.self]
        )
    }
}
```

PR #69, opened 2018-03-31
Fix duplicate bucket name in URL normalization

tengyifei:
in cases where the bucket name is already part of the host, do not put bucket name in front of the host.
e.g. using S3 accelerated transport, the host is always `[bucket].s3-accelerate.amazonaws.com`


tengyifei:
Yeah of course

noppoMan:
merging. Thanks for your contribution😁

PR #70, opened 2018-03-31
Fix #68

### LINKED ISSUE #68: ReplicationStatus.complete should be completed
See https://github.com/noppoMan/aws-sdk-swift/blob/ae8b82371280086cf81c4525331c9b69e860f403/Sources/AWSSDKSwift/Services/s3/S3_Shapes.swift#L1516

But AWS docs uses `COMPLETED`: https://docs.aws.amazon.com/AmazonS3/latest/dev/crr-status.html

Hence under the current code, you can't get objects from any replicated buckets. Doing so will result in a `DecodingError`.

tengyifei:
If this is generated from the AWS official API json, perhaps we should contact them regarding the error?


jonnymacs:
fixed in #88

PR #75, opened 2018-06-27
fix compile error with Swift 4.1 and no longer support Swift 3

noppoMan:


jonnymacs:
@Yasumoto we'll need to setup travis for this org, after which builds should no longer fail after 10 minutes without any output. meanwhile, I will merge this

PR #76, opened 2018-08-06
Sync models with aws-sdk-go@v1.15.5

noppoMan:
@jonnymacs @Yasumoto 
I synchronized entire models with [aws-sdk-go@v1.15.5](https://github.com/aws/aws-sdk-go/tree/v1.15.5/models).

**Brand New Services**
* acm-pca
* autoscaling-plans
* connect
* dlm
* eks
* fms
* iot1click-devices
* iot1click-projects
* iotanalytics
* macie
* mediatailor
* neptune
* pi
* secretsmanager
* servicediscovery
* transcribe
* workmail

**Updated Services**
almost

#### Memo
When updating the models so far, I always brought models from [here](https://github.com/aws/aws-sdk-go/tree/master/models) and replaced it with our models and then run `aws-sdk-swift-codegen` from Xcode or Terminal.
Anyways I'll add guide(About Code generation, publishing separate SPM Packages) for developers asap.


jonnymacs:
Thanks @noppoMan. I know @Yasumoto has some experience with that code-gen process, but I haven't gone through it yet. So yes, a guide will be very helpful.

If yo don't mind I will rebase off master so I can see the build pass :)

Yasumoto:
Thank you so much, @noppoMan!

I think this is a good sign to us that we'll want/need to split out into separate repositories (that are all still coordinated + synced with one process together) so we can gain the benefits of only including the services we need.

Issue #80, opened 2018-08-17
Publishing separate SPM Packages through Github's API


noppoMan:
@jonnymacs @Yasumoto 
Now I'm thinking how we can automate to publish separate SPM Packages(eg. swift-aws/ec2) flow.

My plan is [Publishing separate SPM Packages through Github's API](https://developer.github.com/v3/git/commits/#create-a-commit) as described in title.

The work flow is...

1. Updating models and  running CodeGenerator
2. Make a PR
3. Merging
4. Make a Tag(must follow semantic versioning) and push it to github, Then the script for publishing packages(commit, push, tagging) is executed on Travis(if the tag is followed semantic versioning)

How do you think this?

Yasumoto:
I think this is a pretty great idea. 👏 

An alternative I don't think will work (but I would like to try in parallel just in case) is to keep all of these together, but make a new `Target` for each service. If we were to still include a "global" `AWSSDKSwift` target then that'd still take "forever" to build, but I'm still curious to see the data.

jonnymacs:
I thought about the target approach as well. But then you have to d/l all the other services that you don’t ultimately need. Sure they don’t compile but still a bit of a downer.

The way spm is designed I think we really need to just have a ton of git repos.

Short answer I think Yuki’s approach is the best way to go.

I do want to try and bang out https://github.com/swift-aws/aws-sdk-swift/issues/78 before we build all the repos though. Thoughts?

Yasumoto:
On a relatively beefy EC2 instance:

```
3059.32user 92.78system 4:02.69elapsed 1298%CPU (0avgtext+0avgdata 1691004maxresident)k
```

vs [this branch](https://github.com/swift-aws/aws-sdk-swift/pull/81):

```
1696.18user 62.52system 3:15.81elapsed 898%CPU (0avgtext+0avgdata 1660268maxresident)k
```

I'm curious at the CPU difference, so I'm just going to be bold and make a PR to see what happens to Travis.

jonnymacs:
@noppoMan @Yasumoto fyi I'll be working on #78 this week to unblock this

Yasumoto:
I spent a bit more time looking at #81, and I actually like it a good bit now.

I realized this it's similar to [NIO's approach](https://github.com/apple/swift-nio/blob/master/Package.swift). Even though for _building_ the entire git repository gets pulled along, it doesn't get included in the final product.

[...]

Issue #78, opened 2018-08-13
follow aws naming convention for modules


jonnymacs:
Modules are built off the directory structure of the models/apis/*.json and the metadata['endpointPrefix']

Many services have names that do not match this convention (email for example is SES)

We should follow the pattern as is done in other sdks

https://github.com/aws/aws-sdk-go-v2/blob/996478f06a00c31ee7e7b0c3ac6674ce24ba0120/private/model/api/api.go#L105

PR #81, opened 2018-08-17
Separate modules for all services

Yasumoto:
An investigation for #80 that likely shouldn't be merged.


Yasumoto:
@jonnymacs what're your feelings about this? The tests are failing because the S3 tests themselves are hanging, but I can take a look after we sort out this idea. I think this would be a precursor to your work renaming the modules as we're automatically generating the names, but I don't want to push this through unless you're good with it too 👍

jonnymacs:
@Yasumoto ok yeah, this code builds module names off the directory structure. So should not need any changes once i get #78 finished.

I'm on board with this. Let's move forward with it.

PR #85, opened 2018-09-07
Type issue converting what is a String that looks like an Int

Yasumoto:
@jonnymacs mentioned that we don't utilize the models to do our parsing, but I think they provide important hints on matching the types we'll use for our `Codable` structs.

For example, `Elasticloadbalancing.describeLoadBalancers` returns:

`<SourceSecurityGroup><OwnerAlias>123458855789</OwnerAlias><GroupName>https</GroupName></SourceSecurityGroup>`

which results in:

```
Fatal error: Error raised at top level: Swift.DecodingError.typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "LoadBalancerDescriptions", intValue: nil), _DictionaryKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "SourceSecurityGroup", intValue: nil), CodingKeys(stringValue: "OwnerAlias", intValue: nil)], debugDescription: "Expected to decode String but found Int instead.", underlyingError: nil)): file /home/buildnode/jenkins/workspace/oss-swift-4.1-package-linux-ubuntu-16_04/swift/stdlib/public/core/ErrorType.swift, line 191
```

Maybe it's okay to ship this patch as a temporary workaround, but it's going to be overwritten the next time we generate the code and need to be manually fixed.

Ideas and suggestions welcome! 🙏

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -153,7 +153,7 @@ let package = Package(
 	.library(name: "Xray", targets: ["Xray"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/swift-aws/aws-sdk-swift-core.git", .branch("xml-parsing")),
+        .package(url: "https://github.com/swift-aws/aws-sdk-swift-core.git", .upToNextMajor(from: "2.0.0-rc.1")),
         .package(url: "https://github.com/IBM-Swift/SwiftyJSON.git", .upToNextMajor(from: "17.0.2"))
     ],
     targets: [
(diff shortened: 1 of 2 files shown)


jonnymacs:
thoughts on the above suggestion? I can make the changes for this and #68 and submit a separate pr.

Yasumoto:
This is solved by https://github.com/swift-aws/aws-sdk-swift/pull/88 (though we'll still need some kind of automated/longer term solution)

PR #90, opened 2018-11-27
Fix build error with Swift 4.2.1

noppoMan:
Hi, @jonnymacs @Yasumoto 

Today, I tried to adopt aws-sdk-swift@2.0.0 to [Hexaville](https://github.com/noppoMan/Hexaville).
But I got a following error when I run `swift build` with Swift 4.2.1(macOS)

```
AWSSDKSwift/Sources/AWSSDKSwift/Services/S3/S3_API.swift:6:8: error: no such module 'S3RequestMiddleware'
import S3RequestMiddleware
```
This error is simple,  Cause `S3Middleware` was not defined as dependent on the `S3` target.
So I added `S3Middleware` as one of dependency of `S3` target like `AWSSDKSwift`.

Additionally, I found compile error in unit tests. I also fixed it as well.


jonnymacs:
@noppoMan nice catch

https://github.com/swift-aws/aws-sdk-swift/blob/master/scripts/generate_targets.swift#L41

PR #97, opened 2019-01-14
update core with fix for path in params

jonnymacs:
update nio and swiftJSON to latest versions


jonnymacs:
@Yasumoto objections to my merging this and tagging a release?

Issue #98, opened 2019-01-30
Glacier APIs, always throw exception "Required parameter missing: API version"


adam-fowler:
It seems like all the Glacier code is throwing the same error. The following code 
```
do {
    let glacier = Glacier(region:.euwest2)
    let description = try glacier.describeVault(Glacier.DescribeVaultInput(vaultName:"vaultName", accountId:"-"))
    print(description)
} catch let error {
    print(error)
}
```
Always outputs the following error
```
AWSError(message: "Required parameter missing: API version", rawBody: "{\"code\":\"MissingParameterValueException\",\"message\":\"Required parameter missing: API version\",\"type\":\"Client\"}")
```
I've tried this with listVaults() and listJobs() with the same results

jonnymacs:
Yeah. That’s good thinking.

I am thinking I’d like to avoid a service specific customization in the
core lib directly by implementing a glacier middleware. I have something
that should work, I’ll push up a branch when it’s ready (today / tonight)

jonnymacs:
fixed in #102

Issue #99, opened 2019-02-03
Glacier Job payload should be in JSON but it isn't


adam-fowler:
Whenever I try to send an InitiateJob() request to Glacier, it always returns the following error.
```
AWSError(message: "SerializationException", rawBody: "{\"code\":\"SerializationException\",\"type\":\"Client\",\"message\":\"SerializationException\"}")
```
The cause of this look like it is because the body of the InitiateJob() request contains the JobParameters and these should be in json. But it is coming out in this form
```
JobParameters(retrievalByteRange: nil, tier: nil, format: nil, selectParameters: nil, description: Optional(\"Inventory\"), inventoryRetrievalParameters: Optional(Glacier.Glacier.InventoryRetrievalJobInput(marker: nil, limit: nil, endDate: nil, startDate: nil)), outputLocation: nil, type: Optional(\"inventory-retrieval\"), archiveId: nil, sNSTopic: nil)
```
This looks like it could be a common issue across all AWS systems where a json body is required.

jonnymacs:
Interesting. There are about 90 services that specify the "json" protocol,
one of which is kinesis which I am using the client to post JSON without
any errors.

However another 55 specify "rest-json", one of which is glacier. These
services may be affected by this issue.

adam-fowler:
The JobParameters are output as the payload. And it is the output of the payload that is the problem. The Kinesis payloads look like they are raw data so wouldn't be affected by this issue.

adam-fowler:
I temporarily sorted it by checking if the payload was an AWSShape and if it was then outputting it as JSON. To get that to work was like pulling teeth, I'm not very experienced with the reflection code so wasted a lot of time. I ending up writing the full Input to json, parsing it into a dictionary and then writing the payload element of the dictionary back out to json.

jonnymacs:
ah, ok I get it now. I'll have to ask @noppoMan if how this ever worked?

Could you try

```
switch payloadBody {
    case is AWSShape:
        body = .json(try AWSShapeEncoder().encodeToJSONUTF8Data(payloadBody as! Input))
    default:
        body = Body(anyValue: payloadBody)
}
```

adam-fowler:
> Could you try
> 
> ```
> switch payloadBody {
>     case is AWSShape:
>         body = .json(try AWSShapeEncoder().encodeToJSONUTF8Data(payloadBody as! Input))
>     default:
>         body = Body(anyValue: payloadBody)
> }
> ```
[...]

PR #102, opened 2019-03-01
fixes for Glacier Service

jonnymacs:
- update core with fix to properly serialize the payloadPath objects as json
- add GlacierMiddleware to add apiVersion header, and treeHash checksum headers
- update core nio, and nio-ssl


jonnymacs:
@adam-fowler if you have a chance to test this branch I'd be grateful. This should resolve #98 and #99. 

I also added support for the glacier treeHash checksum so it should not required as input for apis that require it. Not sure if you were needing this or not, but I saw it in the docs and decided to add it

adam-fowler:
Hi Jonathan,

I’m quite busy at the moment so not sure when I’ll get a chance to check
this out. Should be able to before the end of next week though.

Adam

adam-fowler:
Ran the code and it all looks good. The tree hash stuff is probably unnecessary as when it is required a local copy is usually kept. For instance when I do a multipart upload of an archive I need to keep a copy of the tree hash for each part of the archive, so I can construct a tree hash for the full archive which is sent in the completeMultipartUpload() call.

jonnymacs:
🙌

re the tree hash - I'll leave it in there if it doesn't hurt anything since I see the header is added in other sdks (ruby, go)

adam-fowler:
Yeah sounds good

PR #110, opened 2019-03-29
Ensure AWSShape always has a public init function

### LINKED ISSUE #108: STS.GetCallerIdentityRequest initializer is inaccessible
Code generator needs to add a ```public init() {}``` to STS.GetCallerIdentityRequest class

adam-fowler:
Fixes problem with constructing STS.GetCallerIdentityRequest
Fixes #108.

I haven't run the code generator though because it seems to be re-ordering all the code. This isn't too much of a problem except when it is re-ordering the parameters in init() functions. I noticed this has already happened in December in commit https://github.com/swift-aws/aws-sdk-swift/commit/0360367d8cb531a256cde7714cc0033b17fa4490.

I would suggest you force the order to be alphabetical. Then they won't change between builds of code generator and model changes.

### DIFF
--- a/Sources/CodeGenerator/CodeGenerator.swift
+++ b/Sources/CodeGenerator/CodeGenerator.swift
@@ -393,13 +393,13 @@ extension AWSService {
                     code += "\(indt(2))public \(member.toSwiftImmutableMemberSyntax())\n"
                 }
                 code += "\n"
-                if type.members.count > 0 {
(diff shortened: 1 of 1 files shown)


jonnymacs:
@adam-fowler I'm going to work on the re-order issue, then tag the core and regen the code here and tag a rel.

jonnymacs:
This change is in #115

PR #115, opened 2019-04-18
Init param order

jonnymacs:
this PR includes #110


adam-fowler:
I was going to recommend you alphabetize the shapes, API and errors. Means whenever you regenerate you can diff changes with previous code a lot easier.

jonnymacs:
Good idea. I’ll do this as well.

adam-fowler:
This all looks good.

PR #120, opened 2019-05-10
do not populate content-md5 on s3 GET requests

jonnymacs:


adam-fowler:
Looks good, as I said GET requests shouldn't have a body, but we can fix that later

PR #127, opened 2019-06-02
Shell script for updating shapes and api files

adam-fowler:
- Gets models from aws-sdk-go
- Copies them into aws-sdk-swift project
- Builds code generator
- Runs code generator
- Builds service files

Thought we could update the model files as the last build was done 9 months ago. Then thought automating this would make it a lot quicker and easier to keep these up to date.

### DIFF
--- a/scripts/update_models.sh
+++ b/scripts/update_models.sh
@@ -46,7 +46,7 @@ build_files()
 check_for_local_changes()
 {
     LOCAL_CHANGES=`git status --porcelain`
-    if [ -n "$LOCAL_CHANGES" ] && [ $COMMIT_CHANGES=1 ]; then
+    if [ -n "$LOCAL_CHANGES" ] && [ "$COMMIT_CHANGES" == 1 ]; then
         echo "You have local changes already and you have requested to commit changes after this script has run.\nEither remove the -c option or stash your local changes."
         usage
         exit
@@ -102,6 +102,6 @@ copy_model_files $AWS_SDK_GO_MODELS $TARGET_MODELS
 #build service files from the models and check they compile
 build_files
 
-if [ $COMMIT_CHANGES=1 ]; then
+if [ "$COMMIT_CHANGES" == 1 ]; then
     commit_changes
 fi
--- a/scripts/update_models.sh
+++ b/scripts/update_models.sh
@@ -1,33 +1,41 @@
 #!/bin/sh
 
-set -e
+set -eux
+
+TEMP_DIR=""
+COMMIT_CHANGES=""
 
 usage()
 {
-  echo "Usage: update_models.sh -c [ -v MODELS_VERSION_NUMBER ]"
-  exit 2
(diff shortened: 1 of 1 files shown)


adam-fowler:
Added options to 
- specify a aws-sdk-go version number 
- commit changes immediately

scripts/update_models.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/sh
+
+set -e
```
Yasumoto: How about `set -eux` ? 🙏

scripts/update_models.sh
```diff
@@ -0,0 +1,107 @@
...
+
+usage()
+{
+  echo "Usage: update_models.sh -c [ -v MODELS_VERSION_NUMBER ]"
+  exit 2
+}
+
+get_aws_sdk_go()
+{
+  # clone aws-sdk-go into folder
```
Yasumoto: Small nit, any chance you could do something like

```
DESTINATION_FOLDER="$1"
BRANCH_NAME="$2"
```

here and elsewhere to make it clear what the positional args are for? 🙏

scripts/update_models.sh
```diff
@@ -0,0 +1,107 @@
...
+    swift build
+}
+
+check_for_local_changes()
+{
+    LOCAL_CHANGES=`git status --porcelain`
+    if [ -n "$LOCAL_CHANGES" ] && [ "$COMMIT_CHANGES" == 1 ]; then
+        echo "You have local changes already and you have requested to commit changes after this script has run.\nEither remove the -c option or stash your local changes."
+        usage
+        exit
```
Yasumoto: `usage` should already call `exit 2` so this isn't necessary

scripts/update_models.sh
```diff
@@ -0,0 +1,107 @@
...
+    fi
+}
+
+commit_changes()
+{
+    COMMIT_MSG="Sync models with aws-sdk-go "$AWS_MODELS_VERSION
```
Yasumoto: I almost always double-quote my variables in shell, so this should be:

`COMMIT_MSG="Sync models with aws-sdk-go $AWS_MODELS_VERSION"`

scripts/update_models.sh
```diff
@@ -0,0 +1,107 @@
...
+        exit ;;
+    esac
+done
+
+
+trap cleanup EXIT
+
+check_for_local_changes
+
+#create temp folder
```
Yasumoto: I'd suggest removing these, your echo statements do a great job of describing what's going on already.

adam-fowler:
Updated the script with your change requests, ran shellcheck on it and fixed issues.
I'm happy for shellcheck to run on all shell scripts if you want to add it in

adam-fowler:
@jonnymacs @Yasumoto Are you guys happy for me to do a rebuild of all the Shape files?

PR #136, opened 2019-06-17
use swiftyJson from swiftyJson instead of IBM

jonnymacs:


jonnymacs:
I need the Codable support in SwiftJSON 4.1+ for a project that creates a dependency issue with our IBM version.

adam-fowler:
Have you checked the CodeGenerator still produces the same output?

Issue #130, opened 2019-06-04
Run shellcheck on all shell scripts


Yasumoto:
As discussed in #127

adam-fowler:
spellcheck or shellcheck?

Yasumoto:
Darn it autocorrect...

PR #138, opened 2019-06-19
Run shellcheck on our files

Yasumoto:
This is for #130


adam-fowler:
This uses Travis which we don't use anymore, I'm just gonna close this.

PR #146, opened 2019-07-10
move the code generator out of the main project

jonnymacs:
in this way any code gen dependencies are not imposed on the sdks

remove swifty json as an sdk dependency

fix the update_models script to work


adam-fowler:
That all looks good

PR #148, opened 2019-07-16
Deal with situation where location is set to "headers"

adam-fowler:
Some S3 Shapes have a location setting "headers" not "header", this was not being dealt with. 
While this means the variables aren't being dealt with it also was creating another issue where a body is being created for operations that shouldn't have one. 

@jonnymacs Would be good to get this in release 3.1

### DIFF
--- a/CodeGenerator/Sources/CodeGenerator/CodeGenerator.swift
+++ b/CodeGenerator/Sources/CodeGenerator/CodeGenerator.swift
@@ -40,7 +40,7 @@ extension Location {
         case "querystring":
             self = .querystring(locationName: name)
 
-        case "header":
+        case "header", "headers":
             self = .header(locationName: name)
 
         case  "body":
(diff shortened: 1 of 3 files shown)


jonnymacs:
@adam-fowler test case failing on linux

`Test Case 'AWSRequestTests.testS3CreateMultipartUpload' started at 2019-07-16 21:37:37.912
/home/travis/build/swift-aws/aws-sdk-swift/Tests/AWSSDKSwiftTests/AWSRequestTests.swift:26: error: AWSRequestTests.testS3CreateMultipartUpload : XCTAssertNotNil failed - 
Fatal error: Unexpectedly found nil while unwrapping an Optional value`

adam-fowler:
Yep saw that, That test isn't valid anymore as S3.CreateMultipartUpload() doesn't create a body. I've replaced it with a new one that tests a different part of S3.

PR #155, opened 2019-08-01
Add validation code to AWSShapes 

adam-fowler:
This PR is dependent on PR https://github.com/swift-aws/aws-sdk-swift/pull/153.
This PR also requires the core PR https://github.com/swift-aws/aws-sdk-swift-core/pull/102

This adds validate functions to AWSShape classes where required. The validation is based on the min, max and pattern fields of the model files. Where a member is a shape that requires validation, its own validation function is called. Where a member is an array of a shape that requires validation then each element of that array has its validation function called.

I also added auto-generation of idempotency tokens in this PR thus fixing https://github.com/swift-aws/aws-sdk-swift/issues/154


adam-fowler:
Reminder to self, don't merge this until core code is submitted and Package.swift has been edited to point to master branch, or release including core code.

jonnymacs:
just fyi, i see you merged already 👍 

I did a little testing this morning with Kinesis. Looking good!

`"PutRecordsInput.streamName (foo_%^_bar) does not match pattern [a-zA-Z0-9_.-]+."
`
One thing that may not be so useful is the error message when a collection exceeds the max value. The error includes the collection, which will be very long if it exceeds the max value...

`"Length of PutRecordsInput.records [ a very long collection] is greater than the maximum allowed value 500."`

```
    public func validate<T : Collection>(_ value: T, name: String, parent: String, min: Int) throws {
        guard value.count >= min else { throw AWSClientError.validationError(message: "Length of \(parent).\(name) (\(value)) is less than minimum allowed value \(min).") }
    }
    public func validate<T : Collection>(_ value: T, name: String, parent: String, max: Int) throws {
        guard value.count <= max else { throw AWSClientError.validationError(message: "Length of \(parent).\(name) (\(value)) is greater than the maximum allowed value \(max).") }
    }
```

I think these validations should just print the length of the collection not, the value. I can make a PR for this in the core lib if no objections.

adam-fowler:
go ahead

PR #158, opened 2019-08-13
Fix S3.GetBucketLocation by editing response body

adam-fowler:
Fixes https://github.com/swift-aws/aws-sdk-swift/issues/157
This pull request needs https://github.com/swift-aws/aws-sdk-swift-core/pull/106

Also
- renamed AWSRequestMiddleware to AWSServiceMiddleware
- S3 GET requests don't mangle the hostname if it isn't amazonaws.com as is done for all other requests

### DIFF
--- a/Sources/AWSSDKSwift/Middlewares/S3/S3RequestMiddleware.swift
+++ b/Sources/AWSSDKSwift/Middlewares/S3/S3RequestMiddleware.swift
@@ -1,7 +1,7 @@
 import Foundation
 import AWSSDKSwiftCore
 
-public struct S3RequestMiddleware: AWSRequestMiddleware {
+public struct S3RequestMiddleware: AWSServiceMiddleware {
 
     public init () {}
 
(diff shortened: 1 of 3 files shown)


Sources/AWSSDKSwift/Middlewares/S3/S3RequestMiddleware.swift
```diff
@@ -15,6 +15,7 @@ public struct S3RequestMiddleware: AWSRequestMiddleware {
 
         switch request.httpMethod.lowercased() {
         case "get":
+            guard let host = request.url.host, host.contains("amazonaws.com") else { break }
```
jonnymacs: 🤔 on line 21 we check for this, and support when it's not true. Is there a valid use case where the host does not contain `amazonaws`?

jonnymacs: right. So then what is the purpose of the code on lines 21-26 after this change?

adam-fowler: I guess they can be removed, I'll do that

adam-fowler: Removed this change so we can submit this PR. Will add another PR for this

PR #164, opened 2019-08-22
Add "self." in front of validated variables called "name"

adam-fowler:
If we don't do this we end up validating the name variable passed in as a parameter.


CodeGenerator/Templates/struct.stencil
```diff
@@ -36,10 +36,10 @@
         public func validate(name: String) throws {
 {%for v in struct.validation %}
 {%if v.shape %}
-            try {{v.name}}{%if not v.required %}?{%endif %}.validate(name: "\(name).{{v.name}}")
+            try {%if v.name == "name" %}self.{%endif %}{{v.name}}{%if not v.required %}?{%endif %}.validate(name: "\(name).{{v.name}}")
```
jonnymacs: why not just `self.` on all of the fields we are validating. I think this will make it consistent and explicit.

adam-fowler: Err yeah dunno why I didn't do that. Would simplify things. I'll change it

PR #168, opened 2019-09-05
Update standalone module generation process

adam-fowler:
The swift-aws standalone modules are getting out of date. I thought I would look at generating new versions. I tidied up the process a little and made it more robust. 
- Limited the output of main.js to only the modules we want to be standalone
- Added command line parameter to main.js for AWSSDK version to use
- Added additional Package.swift format for packages that have middleware
- Removed AWSSDK prefix from module names to be consistent with modules names in aws-sdk-swift
- publish-module.sh doesn't need folders already setup with git. It will initialise git, add the github remote, reset to state of remote, commit changes, and push to remote.

With these changes the publish process should work without intervention
```
node ./module-exporter/main.js <working folder> <version number>
./module-exporter/publish-module.sh <working folder> <commit comment> <version number>
```


jonnymacs:
@adam-fowler i appreciate this work. but.. my thoughts have been that we don't need stand alone modules anymore.

The stand alone modules were the original solution to speeding up the compile times. The main sdk used to compile each service module and clients had to include the entire thing. 

When we were addressing this there were two proposals, one was to automate the generation of a standalone github module for each service, and the other was the solution we went with to create individual libs / targets for each module. Once this was done these stand along modules were no longer necessary.

Given clients can just include the entire sdk, but then choose to compile only services they want do you have a strong use case for these stand alone modules? If not we should explore how to remove them from the swift-aws namespace on github

adam-fowler:
Hey @jonnymacs I did this work because I was creating a AWS lambda function using Docker. I found whenever I needed to build the lambda function it took forever to build and most of that time seemed to be getting latest on aws-sdk-swift. I thought getting one of the smaller repos might be quicker. I'm not that experienced with Docker so maybe there is a quicker route to get this to work without having to get the aws-sdk-swift repo every time I build.

I guess the other issue is we have these repositories, which still have traffic but are not getting updated. We could update them one last time and then add something to the readme that indicates they won't be updated in the future. You can archive repos which mean they are still available to use but are not available to edit.

Issue #173, opened 2019-09-14
Error on FileUpload to S3


pankajsoni19:
I am following example on readme page of this repo.

Getting:

`NIOOpenSSL.NIOOpenSSLError.unableToValidateCertificate`

adam-fowler:
Which example? How are you building your code? What version of the aws-sdk-swift are you using? What platform are you running on? What are you running this on (ec2, lambda, ...)? I've generally seen this when there is a discrepancy between the ssl lib used to build code and the one the code is run with.

adam-fowler:
Are you running the same setup as in your previous issue?

pankajsoni19:
I am building code using XCode > Version 10.3 (10G8)  [ Latest ]
MacOS > 10.14.6 (18G95) [ Latest ]

Using `master` repo
```
.package(url: "https://github.com/swift-aws/aws-sdk-swift.git", .branch("master")),
```

Upload function

```
 public func upload(_ bodyData: Data, in dir: S3Dir, withKey name: String) throws -> Future<HTTPResponseStatus> {
        let fileKey = dir.s3Path(filename: name)
        let putObjectRequest =
            S3.PutObjectRequest(acl: .publicRead,
                                body: bodyData,
                                bucket: config.awsBucket,
                                contentDisposition: dir.contentDisposition(filename: name),
                                contentLength: Int64(bodyData.count),
                                contentType: dir.contentType,
                                key: fileKey,
                                tagging: "type=\(dir.rawValue)")
        
        return try client.putObject(putObjectRequest).map { (response: S3.PutObjectOutput) in
            return HTTPResponseStatus.ok
        }
    }
```

pankajsoni19:
OpenSSL

```
> which openssl
/usr/bin/openssl

> /usr/bin/openssl version
LibreSSL 2.6.5
```

adam-fowler:
So you are building on Mac and running on the same Mac? Hmmm weird. I’ve been running pretty much similar code this morning fine. 

If you could try v3.1 to see if that works that’d help. Also there is an nio2.0 branch which avoids a lot of these issues as swift nio2.0 has its own version of libssl

adam-fowler:
One other question do you have a custom endpoint?

pankajsoni19:
1. Nope do not have a custom endpoint.
2. Changed to use         
```
.package(url: "https://github.com/swift-aws/aws-sdk-swift.git", from: "3.1.0")
```
same error

[...]

PR #174, opened 2019-09-14
S3 bucket name validation

adam-fowler:
Added ability to add entries to dictionaries in patch code of CodeGenerator
Added regex pattern validation for s3 bucket name.
- Bucket names can have lowercase letters, numbers or hyphen
- Bucket names can only start and end with a lowercase letter or number
- Originally bucket names could include periods but these don't work with s3 virtual addressing so I am not allowing these here.

The user will get a more useful error now instead of the SSL unableToValidateCertificate error you get if you use a bucket name containing a period which is not informative at all. See #173

### DIFF
--- a/CodeGenerator/Sources/CodeGenerator/patch.swift
+++ b/CodeGenerator/Sources/CodeGenerator/patch.swift
@@ -41,6 +41,7 @@ let servicePatches : [String: [Patch]] = [
     "S3": [
         Patch(.replace, entry:["shapes","ReplicationStatus","enum",0], value:"COMPLETED", originalValue:"COMPLETE"),
         Patch(.replace, entry:["shapes","Size","type"], value:"long", originalValue:"integer"),
+        // Add additional location constraints
         Patch(.add, entry:["shapes", "BucketLocationConstraint", "enum"], value:"us-east-2"),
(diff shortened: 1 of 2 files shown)


aymanbagabas:
> * Originally bucket names could include periods but these don't work with s3 virtual addressing so I am not allowing these here.

I'm running into this exact issue. According to [this](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules), periods in s3 virtual addressing is fine but gives ssl trouble. However, a simple solution could be to use this form `https://s3.amazonaws.com/{bucket.with.periods}/{key}`. [StackOverFlow](https://stackoverflow.com/a/24602364)

adam-fowler:
@aymanbagabas Amazon are phasing out that form of addressing s3 buckets in 2020.

adam-fowler:
@Yasumoto I’m wondering if we can’t support both versions of s3 bucket url. Maybe pass a flag to the middleware somehow. I noticed the aws js sdk does this.

adam-fowler:
This has now been reverted in #198. If a bucket has a period in its name we don’t apply virtual addressing.

Issue #177, opened 2019-09-30
Compile failure on Linux Swift 5.1


Andrea-Scuderi:
See https://github.com/swift-aws/aws-sdk-swift-core/issues/123

adam-fowler:
Hey @Andrea-Scuderi this is resolved now yeah?

Andrea-Scuderi:
Yes thanks!

PR #178, opened 2019-09-30
[#177] Support Swift 5.1

Andrea-Scuderi:
Add canImport(FoundationNetworking)


adam-fowler:
Hey @Andrea-Scuderi what is not compiling if you don’t add this. Seems odd our CodeGenerator would need networking code.

Andrea-Scuderi:
@adam-fowler Good spot! I double checked and it's not needed. I'll close the PR.

PR #179, opened 2019-10-11
add support for downloading a file from s3

jonnymacs:
- download a file from s3 to disk in serial 5MB chunks. surely this can be enhanced further,
but this is a start
 - modeled somewhat after the ruby sdk
 - support generation of custom api extensions in the
code generator


adam-fowler:
I've got something similar hanging around. It does a chunked download and upload. Each chunk is either provided by/sent to a callback. There are specialized versions which load from/save to a file.

I created a new module for mine 's3extension'. You can find it here https://github.com/adam-fowler/aws-sdk-swift/blob/s3-extensions/Sources/AWSSDKSwift/Extensions/S3/S3_Extension.swift.

We should combine the two with the best bits. I'll have a proper look at yours over today.

jonnymacs:
ha. ok, note to self - ask @adam-fowler if he's worked in something before doing it :)

I'm not loyal to the generation. In fact I like your approach better as it doesn't require a bunch of code to be maintained in a template file. In either case we'll need to enhance the doc generation to know about the extensions.

Nice catch with the blocking function, and yes this can be resolved by doing the writing async [(like you are doing)](https://github.com/adam-fowler/aws-sdk-swift/blob/s3-extensions/Sources/AWSSDKSwift/Extensions/S3/S3_Extension.swift#L34)

I piggy backed on Data's ability to write out the first bytes to a file directly then extended data to support an append function using FileHandle. Hence the use of both. Since this is a serial operation it seems fine.

Using either FileHandle or OutputStream works. Again, not loyal but as you say FileHandle a little cleaner - you can just pass the response body Data instead of streaming as bytes. And I think we can keep the fileHandle open until the download completes or errors.

There's also c fopen, fwrite... 🤔 

What I like about your Code
- takes a GetObjectRequest as input
- includes multipart upload
- includes S3ErrorType.multipart enum
- includes progress
- takes variable chunkSize as input
- the multipart naming

What I don't like
- no public api?
- the api throws
- returns the fileSize - not sure how useful this is
- generally it's more complicated to read / test / debug. Not that this is so bad, but I do think this is a plug for having an explicit struct to init and execute the various functions. 

Re: hopTo

As I understand things from the [NIO docs](https://github.com/apple/swift-nio/blob/master/Sources/NIO/EventLoopFuture.swift#L343), (and forgive me if I'm misunderstanding but this issue has come up in discord about 6 months back)

The aws client executes calls to aws on it's own eventLoop (https://github.com/swift-aws/aws-sdk-swift-core/blob/master/Sources/AWSSDKSwiftCore/HTTPClient.swift#L162). If these futures are returned to another eventLoop - such a vapor http request connection loop - then preconditionInEventLoop may fail when executing a call back (.then { chunkResponse -> Future<Void> in... )

`hopTo` will ensure callbacks are run on the provided loop and will therefore meet the precondition. NIO also ensures that futures created on that loop will always execute on that loop's thread.

So I *think* it's a good idea to `hopTo` an eventLoop that was created by the existing thread to ensure that precondition is always met.

Looking at your code I think this problem could surface as well.

Conclusion - let's refactor what you have with some of what I have. I'll take a stab at it?

adam-fowler:
Regarding your comments
- no public api
not sure what you mean the functions are public because the extension is public
- the api throws
no it doesn't it only ever returns failed futures. The progress callback sent to the function can be used to cancel the operation midway by throwing an error but this only then returns a failed future.
- returns filesize
yeah not bothered either way with that. 
- harder to read
I can see that to a degree, I think some of that is due to the slightly more complex file handling. I am all for replacing the OutputStream/InputStream with FileHandle code. Lets not go near the c functions.

I'll have to look at the hopTo() documentation more closely, but don't see any harm having them in there.

Do you want me to create a PR with my code so you can edit it?

Oh and my previous PR on core was maybe a little over zealous. Here is a PR to make AWSClient.eventGroup public again https://github.com/swift-aws/aws-sdk-swift-core/pull/126

jonnymacs:
- extension is public. Missed that. Ok 👌 
- doesn’t throw. Another misread. The closures throw and generate failed futures.👌 again. 
- maybe these two reviewer errors are indicative of the code being a little more complicated. 

Yeah, make a pr and we’ll work off it and get it added. Probably looking at a few changes for nio2 also, so I’d say make that a priority then we can backport to our 3.x

adam-fowler:
PR #183 created.

jonnymacs:
@adam-fowler have another look. I wasn't sure how to submit so I kept it here.

I kept the template generation over the extension. I didn't think I'd mind having to add the extension, but I did. I think the multipart api should be available on the default client without having to add the extension in spm or in code.

I refactored to use FileManager / FileHandle. I tested, it works well. I made a few changes you may notice - Data is now processed after the download instead of as input to the next call to mpDlPart.

I added the hop(to: EventLoop) jik.

I updated for Nio2 - will backport to 3.x if this looks good.

I did not test upload... I will before we merge, but let me know what you think first.

adam-fowler:
Extension or Template neither is ideal. We should see if we can work out something better but in the meantime I'm happy to go with the template. From a development point of view the template is quite awkward but it provides a better end user experience.

Comments on the latest code. 
- If you are going to include an on eventloop parameter in one version of the function you might as well do it in the other function (but default it to AWSClient.eventLoop.next() maybe). Of course the same should be done for the upload versions as well.
- I think by moving the data save to after the data download we have lost the parallelism where data blocks are saved to disk while the next block is downloaded from S3. 
- Do you need to call FileManager.default.createFile() before calling FileHandle(forWritingAtPath: filename)?
- We should add an S3 test that uploads and downloads a file to ensure this works?

CodeGenerator/Templates/Custom/s3/multipart_API.stencil
```diff
@@ -0,0 +1,250 @@
...
+    ///     - input: The GetObjectRequest shape that contains the details of the object request.
+    ///     - partSize: Size of each part to be downloaded
+    ///     - filename: Filename to save download to
+    ///     - progress: Callback that returns the progress of the download. It is called after each part is downloaded with a value between 0.0 and 1.0 indicating how far the download is complete (1.0 meaning finished).
+    /// - returns: A future that will receive the complete file size once the multipart download has finished.
+    func multipartDownload(_ input: GetObjectRequest, partSize: Int = 5*1024*1024, filename: String, progress: @escaping (Double) throws -> () = { _ in }) -> Future<Int64> {
+        let eventLoop = AWSClient.eventGroup.next()
+        return eventLoop.submit {  () -> Foundation.FileHandle in
+            if FileManager.default.fileExists(atPath: filename) {
+                try FileManager.default.removeItem(atPath: filename)
```
adam-fowler: The FileManager.default.removeItem() is unnecesary. The subsequent createFile will clear the contents of the file.

CodeGenerator/Templates/Custom/s3/multipart_API.stencil
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /// Multipart upload of file to S3.
+    ///
+    /// - parameters:
+    ///     - input: The CreateMultipartUploadRequest structure that contains the details about the upload
+    ///     - inputStream: The function supplying the data parts to the multipartUpload. Each parts must be at least 5MB in size expect the last one which has no size limit.
+    /// - returns: A Future that will receive a CompleteMultipartUploadOutput once the multipart upload has finished.
+    func multipartUpload(_ input: CreateMultipartUploadRequest, inputStream: @escaping () throws -> Data?) -> Future<CompleteMultipartUploadOutput> {
+        var completedParts: [S3.CompletedPart] = []
```
adam-fowler: Obviously the FileHandle and eventloop changes you made for multipartDownload() will have to be implemented for the upload functions as well.

CodeGenerator/Templates/Custom/s3/multipart_API.stencil
```diff
@@ -0,0 +1,250 @@
...
+        let eventLoop = AWSClient.eventGroup.next()
+        return eventLoop.submit {  () -> Foundation.FileHandle in
+            if FileManager.default.fileExists(atPath: filename) {
+                try FileManager.default.removeItem(atPath: filename)
+            }
+
+            FileManager.default.createFile(atPath: filename, contents: nil)
+
+            guard let fileHandle = FileHandle(forWritingAtPath: filename) else {
+                throw S3ErrorType.multipart.failedToOpen(file: filename)
```
adam-fowler: You should use the FileHandle(forWritingTo: URL) version of the function. It throws an error which could be more informative than failedToOpen()

CodeGenerator/Templates/Custom/s3/multipart_API.stencil
```diff
@@ -0,0 +1,250 @@
...
+public extension S3 {
+
+    /// Multipart download of a file from S3.
+    ///
+    /// - parameters:
+    ///     - input: The GetObjectRequest shape that contains the details of the object request.
+    ///     - partSize: Size of each part to be downloaded
+    ///     - outputStream: Function to be called for each downloaded part. Called with data block and file size
+    /// - returns: A future that will receive the complete file size once the multipart download has finished.
+    func multipartDownload(_ input: GetObjectRequest, partSize: Int = 5*1024*1024, on eventLoop: EventLoop, outputStream: @escaping (Data, Int64) throws -> ()) -> Future<Int64> {
```
adam-fowler: Minor thing, Can you update the function header comment to include the on eventLoop parameter as these are passed into the documentation

CodeGenerator/Sources/CodeGenerator/main.swift
```diff
@@ -52,6 +52,18 @@ for index in 0..<apis.count {
                     encoding: .utf8
                 )
             }
+
+            let customTemplates = service.getCustomTemplates()
+            if !customTemplates.isEmpty {
```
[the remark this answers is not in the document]
jonnymacs: it's not.

jonnymacs:
@adam-fowler I think I addressed everything in your comments, including the parallelism and test coverage. Have another look when you get a chance.

adam-fowler:
All looks good. One minor thing, maybe call `prevPart` `prevPartSave` to make it clearer what it is doing

Issue #187, opened 2019-10-22
Need to support "error", "code" in model files


adam-fowler:
**Describe the bug**
It appears we currently don't support the "code" parameter in the error section of shapes in the model files. We just assume the error code is the same as the error shape's name.

**To Reproduce**
IAM makes extensive use of this parameter and the errors AWS generates are not being recognised.

adam-fowler:
fixed

PR #188, opened 2019-10-22
Extract "error","code" from service model files

adam-fowler:
Some services error shapes have an "error" section that contains a "code" string. This is the string returned by AWS when that error occurs, not the error shape name which was previously used.

This resolves issue #187


Yasumoto:
Kinda feeling like some extra testing would be worthy here, but not sure what beyond some testing by hand. I won't have time to kick the tires on it for another ~2 weeks, but just wanted to share where I got when I took a look.

adam-fowler:
@Yasumoto so far this has resolved issues in IAM and SNS. I haven't tested other services but comparing the documentation for a few of them indicates the changes are valid. As always with these large scale system wide changes it is very hard to test for everything. Hopefully we are getting to the point where there will be less and less of these.

PR #193, opened 2019-11-10
Event loop group provider changes

adam-fowler:
- Added eventLoopGroupProvider parameter to all service client initialisers
- Fixed S3 multipart code to work with changes. I removed the `on eventLoop` parameter from functions as we can provide custom eventLoopGroup at client initialisation


adam-fowler:
removing as Fabien has one prepared already

PR #198, opened 2019-11-14
Don't apply virtual address to s3 buckets contain a period

adam-fowler:
If S3 bucket name contains a period don't apply virutal address process to them as this creates an invalid https address.
Also removed validation not allowing bucket names with a period

### DIFF
--- a/Sources/AWSSDKSwift/Services/S3/S3_Shapes.swift
+++ b/Sources/AWSSDKSwift/Services/S3/S3_Shapes.swift
@@ -62,7 +62,6 @@ extension S3 {
         }
 
         public func validate(name: String) throws {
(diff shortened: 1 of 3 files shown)


Sources/AWSSDKSwift/Services/S3/S3_Shapes.swift
```diff
@@ -180,7 +179,6 @@ extension S3 {
 
         public func validate(name: String) throws {
             try self.filter?.validate(name: "\(name).filter")
-            try self.storageClassAnalysis.validate(name: "\(name).storageClassAnalysis")
```
jonnymacs: how does removing the bucket name pattern validation patch caused this validation to be remove?

adam-fowler: Becuase the only thing inside storageClassAnalysis that needs validating is a bucket name. You have to dig down about 3 or 4 classes before finding it.

PR #203, opened 2019-11-26
Use Localstack to test AWS services

adam-fowler:
Instead of using fakes3 and dynamodb-local use [LocalStack](https://github.com/localstack/localstack) for testing. It provides much more comprehensive versions of S3 and DynamoDB, plus provides local version of a number of other services. 
Changes include
- Use localstack docker image
- Add additional S3 tests
- Add tests for additional services SNS, SQS, IAM, SSM, APIGateway

In addition to including new tests, each set of tests is now setup to not clash with each other. Each system has a TestData object which is created with the name of the test function. This can then be used to create unique test objects (eg buckets, queues) and not worry about another test trampling all over its own assets. The TestData object is expected to clean up after itself.


Tests/AWSSDKSwiftTests/Services/SSM/SSMTests.swift
```diff
@@ -0,0 +1,79 @@
...
+
+// testing json service
+
+class SSMTests: XCTestCase {
+
+    let client = SSM(
+            accessKeyId: "key",
+            secretAccessKey: "secret",
+            region: .apnortheast1,
+            endpoint: "http://localhost:4583",
```
fabianfett: Do we maybe want to expose the host and port as `SSM_HOST` and `SSM_PORT` with fallbacks to `localhost` and `4583`, which would ensure local testability?

That would be great for my pr.

Would apply to all services.

adam-fowler: Do you mean setup an environment variable for each service? Might as well combine the host and the port.

fabianfett: > Do you mean setup an environment variable for each service?
Yes since, we have different ports and so on. And maybe one want's to test locally against dynamo-local. Whatever.

> Might as well combine the host and the port.
That's true. Go ahead with that if you like!

Tests/AWSSDKSwiftTests/test.swift
```diff
@@ -0,0 +1,25 @@
...
+// helper functions for tests
+//
+import XCTest
+import NIO
+import AWSSDKSwiftCore
+
+func attempt(function : () throws -> ()) {
+    do {
+        try function()
+    } catch let error as AWSErrorType {
```
fabianfett: I'm not sure, but I think you can drop the `let error as`

```
} catch AWSErrorType {

}
```
should be sufficient.

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
```diff
@@ -0,0 +1,134 @@
...
+            let testData = try TestData(#function, client: client)
+
+            let messageBody = "!@#$%^&*()-=_+[]{};':\",.<>\\|`~éñà"
+            let sendMessageRequest = SQS.SendMessageRequest(messageBody: messageBody, queueUrl: testData.queueUrl)
+            let messageId = try client.sendMessage(sendMessageRequest).wait().messageId
+            
+            // receive message tests the flattened arrays in XML response
+            var foundMessage = false
+            var messages : [SQS.Message] = []
+            repeat {
```
fabianfett: why do we loop here? We just expect one message?

```
messages.count > 0
```

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
```diff
@@ -0,0 +1,134 @@
...
+                let receiveMessageRequest = SQS.ReceiveMessageRequest(maxNumberOfMessages: 10, queueUrl: testData.queueUrl)
+                messages = try client.receiveMessage(receiveMessageRequest).wait().messages ?? []
+                for message in messages {
```
fabianfett:
```swift
XCTAssertEqual(messages.count, 1)
guard let message = messages.first else { XCTFail("expected to get the first message") }

XCTAssertEqual(message.body, messageBody)

if let receiptHandle = message.receiptHandle {
  let deleteRequest = SQS.DeleteMessageRequest(queueUrl: testData.queueUrl, receiptHandle: receiptHandle)
  try client.deleteMessage(deleteRequest).wait()
}
else {
  XCTFail("expected to have a receipt handle")
}
```

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
```diff
@@ -0,0 +1,134 @@
...
```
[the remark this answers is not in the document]
adam-fowler: Any reason forEach{} is preferable?

Tests/AWSSDKSwiftTests/test.swift
adam-fowler: `catch AWSErrorType` doesn't compile as `AWSErrorType` is an error type. You would need `catch is AWSErrorType` but that doesn't work in this situation because the variable `error` is not setup in this situation. So you do need the `let error as` unless there is another way to do this.

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
adam-fowler: Think where I took this code from was running on a queue that had multiple functions posting to the same queue. I guess in this situation we can get rid of the loop

Tests/AWSSDKSwiftTests/Services/Dynamodb/DynamoDBTests.swift
```diff
@@ -11,76 +11,83 @@ import Dispatch
...
-                "hashKey": DynamoDB.AttributeValue(s: "hello"),
-                "rangeKey": DynamoDB.AttributeValue(s: "world")
-            ],
-            tableName: tableName
-        )
-        _ = try client.putItem(putItemInput).wait()
-    }
+            let createTableInput = DynamoDB.CreateTableInput(
+                attributeDefinitions: [
+                    DynamoDB.AttributeDefinition(attributeName: "hashKey", attributeType: .s),
```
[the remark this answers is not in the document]
adam-fowler: done

Tests/AWSSDKSwiftTests/Services/SNS/SNSTests.swift
```diff
@@ -0,0 +1,95 @@
...
+            guard let topicArn = response.topicArn else {throw SNSTestsError.noTopicArn}
+            
+            self.topicArn = topicArn
+        }
+        
+        deinit {
+            attempt {
+                // createTopic() doesn't appear to create a valid topicArn so cannot delete it
+                //let request = SNS.DeleteTopicInput(topicArn: self.topicArn)
+                //_ = try client.deleteTopic(request).wait()
```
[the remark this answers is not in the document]
adam-fowler: use #if true. 
Localstack doesn't seem to return valid topic arns

Tests/AWSSDKSwiftTests/Services/SNS/SNSTests.swift
```diff
@@ -0,0 +1,95 @@
...
+            let testData = try TestData(#function, client: client)
+
+            let request = SNS.ListTopicsInput()
+            let response = try client.listTopics(request).wait()
+            let topic = response.topics?.first {$0.topicArn == testData.topicArn }
+            XCTAssertNotNil(topic)
+        }
+    }
+    
+    /* commented out until we get valid topic arn's returned
```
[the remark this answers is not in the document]
adam-fowler: done

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
```diff
@@ -0,0 +1,134 @@
...
+                    if messageId == message.messageId {
```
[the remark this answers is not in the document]
adam-fowler: got rid of loop so invalid

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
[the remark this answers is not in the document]
adam-fowler: got rid of loop so invalid

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
```diff
@@ -0,0 +1,134 @@
...
```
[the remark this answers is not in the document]
adam-fowler: got rid of loop so invalid

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
adam-fowler: Removed the loop

Tests/AWSSDKSwiftTests/Services/SQS/SQSTests.swift
adam-fowler: slight difference, I used XCTUnwrap

fabianfett: Didn't know about `XCTUnwrap` that looks great!

Tests/AWSSDKSwiftTests/test.swift
fabianfett: you're right

adam-fowler:
Yeah only found out about XCTUnwrap() a couple of weeks back

PR #204, opened 2019-11-28
GitHub Actions

fabianfett:


adam-fowler:
@fabianfett To get the docker image to see the ports, try adding container option '--network=host'

fabianfett:
@adam-fowler If that's okay for you I would introduce two env variables `DYNAMO_ENDPOINT` and `S3_ENDPOINT` in the tests that default to `localhost`. In docker those endpoints are now exposed as `dynamodb` and `s3`. I would than set the env variables to those endpoints. Nice side effect: You could even run the tests against real endpoints.

adam-fowler:
While you are writing this you should look at the localstack PR. This replaces fakes3 and dynamodb-local.

adam-fowler:
> @adam-fowler If that's okay for you I would introduce two env variables `DYNAMO_ENDPOINT` and `S3_ENDPOINT` in the tests that default to `localhost`. In docker those endpoints are now exposed as `dynamodb` and `s3`. I would than set the env variables to those endpoints. Nice side effect: You could even run the tests against real endpoints.

That'd be fine but be sure we can run tests locally. 

It could have the added advantage of us being able to test on EC2 if the endpoint is set to nil and we remove the credentials. I would get the action running first. Maybe look at this after.

fabianfett:
On hold. Waiting for rebase on landed #203.

adam-fowler:
Added environment variables https://github.com/adam-fowler/aws-sdk-swift/commit/9d9c721417b36aed06b259498659d37e9d656468

adam-fowler:
Hey Fabian, the issues with Linux tests not running was my fault. I removed the key and secret from the client initialisation in each test which broke Linux. It would search for credentials from ec2 instance data and fail. Merge latest localstack and that should fix your issues. Sorry

fabianfett:
@adam-fowler thanks for reaching out. Will fix.

PR #210, opened 2019-12-08
Github Actions: Generate documentation on release

adam-fowler:
Use a github action to generate documentation whenever a new release is published.

Made a couple of changes to the process while doing this.
- Split `gen-doc.sh` into two files `build-doc.sh` and `commit-doc.sh`
- Github action runs `build-doc.sh`
- Make use of latest jazzy which allows multiple sourcekitten files as input
- Include some AWSSDKSwiftCore documentation in the docs.


adam-fowler:
We have one problem with this PR at the moment. The process takes more than an hour and the GITHUB_TOKEN expires after one hour

adam-fowler:
Sorted the issue with the process taking too long by running the sourcekitten json generation across 8 instances

.github/workflows/gen-docs.yml
```diff
@@ -0,0 +1,105 @@
+name: Generate Documentation
+
+on:
+  release:
+    types: [published]
+    branches: "master"
```
fabianfett: We do release from other branches, don't we?

.github/workflows/gen-docs.yml
```diff
@@ -0,0 +1,105 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: Install
+      run: |
+        gem install jazzy
+        npm install lodash fs globby
+    - name: Chunk 1
```
fabianfett: Maybe name should be zero based as well, would make reading easier.

.github/workflows/gen-docs.yml
adam-fowler: We do, but I wanted to get master working first.

.github/workflows/gen-docs.yml
adam-fowler: not sure it makes a huge difference

adam-fowler: > How long would the gen-docs process take if it wouldn't be parallelized?

It takes 75-80 mins which is too long as the GITHUB_TOKEN expires after an hour. Unfortunately when we split across multiple instances we have to re-build the whole of aws-sdk-swift on each instance.

fabianfett: > It takes 75-80 mins which is too long as the GITHUB_TOKEN expires after an hour. Unfortunately when we split across multiple instances we have to re-build the whole of aws-sdk-swift on each instance.

Maybe we should add this as a comment in the yaml file will make it much clearer for future us.

adam-fowler:
Yeah makes sense

.github/workflows/gen-docs.yml
adam-fowler: done

Yasumoto:
This is awesome 👏

PR #214, opened 2020-01-15
Remove XCTUnwrap usage as it is not available in swift 5.0

adam-fowler:
Add Swift 5.0 in continuous integration


adam-fowler:
I had thought of that but seeing as it was one use case I just removed them.

Issue #218, opened 2020-01-16
Prefix all Service Libraries with `AWS`


fabianfett:
Original title: 
Namespaces don't work if Facade struct has same name as Library

Origin of the issue:
https://github.com/fabianfett/swift-lambda-runtime/pull/25

The code that breaks

```swift
import NIO
import DynamoDB
import AWSSDKSwiftCore
import LambdaRuntime

func configureDynamoDB(group: EventLoopGroup, environment: Environment) -> DynamoDB {
    return DynamoDB(
         accessKeyId: environment.accessKeyId,
         secretAccessKey: environment.secretAccessKey,
         sessionToken: environment.sessionToken,
         region: Region(rawValue: environment.region),
         eventLoopGroupProvider: .shared(group)
    )
}
```

`LambdaRuntime` must be imported because of the `Environment` type. Since `LambdaRuntime` also holds a `DynamoDB` type, the compiler emits an error:

```
'DynamoDB' is ambiguous for type lookup in this context
```

### LambdaRuntime Issue

Since the package `LambdaRuntime` has the same name as the class `LambdaRuntime.DynamoDB` the compiler can not differentiate between the namespace and the class. Trying to access: `LambdaRuntime.DynamoDB` leads to the following error: 

```
'DynamoDB' is not a member type of 'LambdaRuntime'
```

For that reason, I will rename `LambdaRuntime` the class to `Runtime`. With this `LambdaRuntime` always refers to the package/namespace and not a struct. This is already in progress with:

https://github.com/fabianfett/swift-lambda-runtime/pull/26

With this fix done, we can now access `LambdaRuntime.DynamoDB` without problems.

### AWSSDKSwift Issue

Even though we can specify `LambdaRuntime.DynamoDB` we still can not specify `DynamoDB.DynamoDB` since we still can't access the namespace. The whole Swift namespace is broken as long as a type has the same name as the library. IMHO we should start a rename here to prevent those naming collisions where one can not fallback on the namespaces. I know this is breaking. But it's also a great feature to not fall apart in the moment of naming collisions.

```swift
import NIO
import DynamoDB
import AWSSDKSwiftCore
import LambdaRuntime

func configureDynamoDB(group: EventLoopGroup, environment: Environment) -> DynamoDB {
    return DynamoDB(
         accessKeyId: environment.accessKeyId,
         secretAccessKey: environment.secretAccessKey,
         sessionToken: environment.sessionToken,
[...]

PR #225, opened 2020-01-19
Prefix all libraries with AWS

adam-fowler:
Prefix all libraries with AWS
See #218

Also added full Package.swift generation using scripts/generate-package.swift


CodeGenerator/Sources/CodeGenerator/patch.swift
```diff
@@ -12,10 +12,6 @@ import SwiftyJSON
 // Patch operations
 //
 let servicePatches : [String: [Patch]] = [
-    "Backup" : [
-        // Backup clashes with a macOS framework, so need to rename the framework
-        Patch(.replace, entry:["serviceName"], value:"AWSBackup", originalValue:"Backup")
```
Yasumoto: This precursor patch (and the one below) likely points out we should've done this all along. Thank you!

PR #227, opened 2020-01-20
Paginate

adam-fowler:
Service code generation for pagination of AWS results. See https://github.com/swift-aws/aws-sdk-swift-core/pull/181 for related PR in aws-sdk-swift-core.
- reads paginators-1.json files from models directory.
- generates {service}_Paginator.swift files. These contain the api calls plus extensions to Input shapes so they can be recreated with a new token.


adam-fowler:
Currently it works for paginators that only expect to pass one token about and only return one results array.

adam-fowler:
closing this in favour of #232. Will bring 4.x.x changes across after that is merged

PR #232, opened 2020-01-23
Paginate into 4.x.x

adam-fowler:
Service code generation for pagination of AWS results. See swift-aws/aws-sdk-swift-core#181 for related PR in aws-sdk-swift-core.

- reads paginators-1.json files from models directory.
- generates {service}_Paginator.swift files. These contain the api calls plus extensions to Input shapes so they can be recreated with a new token.

This drags in a cloudfront fix as well


CodeGenerator/Sources/CodeGenerator/main.swift
```diff
@@ -4,9 +4,25 @@ import Dispatch
...
+        return true
+    }
+}
+
+
 let startTime = Date()
-let apis = try loadAPIJSONList()
-let docs = try loadDocJSONList()
+let models = try loadModelJSON()
+/*let apis = try loadAPIJSONList()
```
fabianfett: can this be removed?

CodeGenerator/Sources/CodeGenerator/main.swift
```diff
@@ -16,63 +32,78 @@ var errorShapeMap: [String: String] = [:]
...
 let environment = Environment(loader: fsLoader)
 
-for index in 0..<apis.count {
-    let api = apis[index]
-    let doc = docs[index]
+for model in models {
 
     group.enter()
+
     DispatchQueue.global().async {
```
fabianfett: we should `defer group.leave()` in here.

CodeGenerator/Sources/CodeGenerator/main.swift
```diff
@@ -16,63 +32,78 @@ var errorShapeMap: [String: String] = [:]
...
         do {
```
fabianfett: Do we maybe want to extract the `do {} catch` block into a throwing function?

CodeGenerator/Sources/CodeGenerator/main.swift
```diff
@@ -4,9 +4,25 @@ import Dispatch
 import Stencil
 import PathKit
 
+extension String {
+    func writeIfChanged(toFile: String, atomically: Bool, encoding: String.Encoding) throws -> Bool {
+        do {
+            let original = try String(contentsOfFile: toFile)
+            guard original != self else { return false }
```
fabianfett: We read the complete file before we write it? I'd assume that this should be far more expensive than just rewriting the file.

CodeGenerator/Sources/CodeGenerator/CodeGenerator.swift
```diff
@@ -499,6 +517,87 @@ extension AWSService {
...
+
+        var paginatorContexts: [PaginatorContext] = []
+        
+        for paginator in paginators {
+            // get related operation and its input and output shapes
+            guard let operation = operations.first(where: {$0.name == paginator.methodName}) else { continue }
+            guard let inputShape = operation.inputShape else { continue }
+            guard let outputShape = operation.outputShape else { continue }
+            guard case .structure(let inputShapeStruct) = inputShape.type else { continue }
+            guard case .structure(let outputShapeStruct) = outputShape.type else { continue }
```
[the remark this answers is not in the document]
adam-fowler: It was easier to debug this way

CodeGenerator/Sources/CodeGenerator/main.swift
adam-fowler: The writeIfChanged() has nothing to do with the performance of the CodeGenerator. It is to do with the effect it has on the XCode build. If I write a file, regardless of whether it changes, the build needs to recompile it. Also XCode does stuff when you change files from underneath it which can have a major performance hit of your Mac. 

As an aside the load to check it it has changed has a minor hit on performance. 

I’ll add a comment detailing this.

adam-fowler: The writeIfChanged() has nothing to do with the performance of the CodeGenerator. It is to do with the effect it has on the XCode build. If I write a file, regardless of whether it changes, the build needs to recompile it. Also XCode does stuff when you change files from underneath it which can have a major performance hit of your Mac. 

As an aside the load to check it it has changed has a minor hit on performance. 

I’ll add a comment detailing this.

CodeGenerator/Sources/CodeGenerator/main.swift
adam-fowler: Ok

CodeGenerator/Sources/CodeGenerator/main.swift
adam-fowler: Happy to do that. You’re looking at inherited code now

Package.swift
```diff
@@ -230,7 +230,7 @@ let package = Package(
         .library(name: "XRay", targets: ["XRay"])
     ],
     dependencies: [
-        .package(url: "https://github.com/swift-aws/aws-sdk-swift-core.git", .upToNextMinor(from: "4.1.0"))
+        .package(url: "https://github.com/swift-aws/aws-sdk-swift-core.git", .branch("paginate"))
```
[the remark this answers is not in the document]
adam-fowler: Yep. Good to have a reminder

CodeGenerator/Sources/CodeGenerator/main.swift
adam-fowler: Could do that. The main.swift could probably have a bit of cleanup.

PR #246, opened 2020-03-02
Fix documentation link in README

0xTim:


adam-fowler:
cheers Tim, I think there might be a few more as well though. Will search them out

PR #268, opened 2020-04-14
Support for streaming response payload

adam-fowler:
The sibling PR for this is https://github.com/swift-aws/aws-sdk-swift-core/pull/236

Code generator outputs additional versions of "streaming" commands that supports passing the streamed payload to a supplied closure. These function have the suffix `Streaming` eg
```
public func getObjectStreaming(_ input: GetObjectRequest, _ stream: @escaping (ByteBuffer, EventLoop)->EventLoopFuture<Void>) -> EventLoopFuture<GetObjectOutput> {
        return client.send(operation: "GetObject", path: "/{Bucket}/{Key+}", httpMethod: "GET", input: input, stream: stream)
    }
```

Added an S3 extension `S3.selectObjectContentEventStream` and extended `S3.SelectObjectContentEventStream` to conform to `AWSClientStreamable` so we can support `selectObjectContent` event streams.


pablogeek:
It looks good to me :) Can we merge?

PR #271, opened 2020-04-16
Use NIONonBlockingFileIO to load/save files in multipart upload/download

adam-fowler:
This will fix the issue where we are blocking the eventloop with file IO. 
- We need a thread pool to run the file IO off, so I have added a threadPoolProvider enum where the user can specify if they want to create one or provide one.
- Now that payload object is available for aws requests we can pass the `ByteBuffer` we get from FileIO directly to aws-sdk-swift. Unfortunately we don't have the same available for responses yet so need to create a new buffer to save out to fileIO which is a little annoying.


pablogeek:
shall we merge this one? I'm looking forward this to work 😉

adam-fowler:
was waiting on a review first from one of the other guys. I'll give them a poke see if I can get a response.

adam-fowler:
> @adam-fowler backport to 3.x?

This would be a breaking change for the v3.x functions as the function signature has changed. But I could add new versions of the functions that use `NonBlockingFileIO`

PR #305, opened 2020-06-02
Fix response metadata middleware

bamx23:
RFC of HTTP headers says that they are case insensitive so it would be nice to allow both x-amz-meta- and X-Amz-Meta- to be valid prefixes in this middleware.


adam-fowler:
This will be fixed in 5.x.x when we start using NIO HTTPHeaders but in the meantime I'm happy for this to go in

bamx23:
Can I expect this to be released in 4.6.1?

adam-fowler:
I've a couple of other changes in aws-sdk-swift-core, which will push the version to 4.7
Also I'll do a model update to get latest service code

adam-fowler:
v4.7.0 is out, both in aws-sdk-swift and s3

bamx23:
That was quick! Thank you

PR #319, opened 2020-06-19
Add Codable support to DynamoDB

adam-fowler:
Use Codable to encode/decode [String: DynamoDB.AttributeValue] dictionaries.
Add extensions to DynamoDB for putIem, getItem, query, scan and updateItem that use the encoder and decoder.


Andrea-Scuderi:
It's really nice! I love it!
It simplify the client code a lot!
I fixed my example:
- before: https://github.com/swift-sprinter/aws-serverless-swift-api-template.git
- after: https://github.com/swift-sprinter/aws-serverless-swift-api-template.git

`scanCodable` is missing? Is it intentional?

adam-fowler:
> `scanCodable` is missing? Is it intentional?

@Andrea-Scuderi  Its good to hear you like it. I don't have a great deal of experience with DynamoDB so wasn't totally sure what I should support while still trying to keep everything fairly unopinionated. I can add a `scanCodable`. Would like to add some form of update but that might need a bit more work.

adam-fowler:
@Andrea-Scuderi added scan and update

Sources/AWSSDKSwift/Extensions/DynamoDB/DynamoDB+Codable.swift
```diff
@@ -0,0 +1,273 @@
...
+            // construct expression attribute name and value arrays from name attribute value map
+            let expressionAttributeNames: [String: String] = .init( item.keys.map { ("#\($0)", $0) }) { first,_ in return first }
+            let expressionAttributeValues: [String: AttributeValue] = .init( item.map { (":\($0.key)", $0.value) }) { first,_ in return first }
+            // construct update expression, if one if not already supplied
+            let updateExpression: String
+            if let inputUpdateExpression = input.updateExpression {
+                updateExpression = inputUpdateExpression
+            } else {
+                let expressions = item.keys.map {"#\($0) = :\($0)"}
+                updateExpression = "SET \(expressions.joined(separator: ","))"
```
Andrea-Scuderi, line 71: This is quite opinionated.

Sources/AWSSDKSwift/Extensions/DynamoDB/DynamoDB+Codable.swift
```diff
@@ -0,0 +1,273 @@
...
+        do {
+            var item = try DynamoDBEncoder().encode(input.updateItem)
+            // extract key from input object
+            var key: [String: AttributeValue] = [:]
+            input.key.forEach {
+                key[$0] = item[$0]!
+                item[$0] = nil
+            }
+            // construct expression attribute name and value arrays from name attribute value map
+            let expressionAttributeNames: [String: String] = .init( item.keys.map { ("#\($0)", $0) }) { first,_ in return first }
```
Andrea-Scuderi, line 63: `expressionAttributeNames` could be the default property of UpdateItemCodableInput

Sources/AWSSDKSwift/Extensions/DynamoDB/DynamoDB+Codable.swift
```diff
@@ -0,0 +1,273 @@
...
+            let expressionAttributeValues: [String: AttributeValue] = .init( item.map { (":\($0.key)", $0.value) }) { first,_ in return first }
```
Andrea-Scuderi, line 64: `expressionAttributeValues` could be the default property of UpdateItemCodableInput

Sources/AWSSDKSwift/Extensions/DynamoDB/DynamoDB+Codable.swift
adam-fowler: If the user supplies no `updateExpression` it is set to `SET...`. So the user can override this

Sources/AWSSDKSwift/Extensions/DynamoDB/DynamoDB+Codable.swift
adam-fowler: see above

Sources/AWSSDKSwift/Extensions/DynamoDB/DynamoDB+Codable.swift
adam-fowler: What would the user setup here apart from the variables taken from the update object?

adam-fowler: I have provided the option to supply your own `expressionAttributeNames` but this is only of any use when the user is supplying their own `updateExpression`.

Sources/AWSSDKSwift/Extensions/DynamoDB/DynamoDB+Codable.swift
adam-fowler: Not sure why you would want this, kinda removes the requirement for the supplied Codable `updateItem`

adam-fowler:
Thinking of removing the word `Codable` from the function names. Function parameters ensure all functions have a different signature from the non-codable versions. It makes the functions slightly more discoverable with code complete. 

Thoughts?

PR #376, opened 2020-09-15
Proof read of the Soto 5 docs

0xTim:
Plenty of things for discussion.

Do we Americanize the docs? That's a habit I've picked up from writing the Vapor book.

Using the same `AWSClient` across the application and all request threads might cause issues since it's a class, so it will be the same instance across all. The services should get a copy since they're `struct`s but the 'Vapor way' for other stuff is to create a new instance for each request and the application


adam-fowler:
> Do we Americanize the docs? That's a habit I've picked up from writing the Vapor book.
> 
I'm ok with Americanization

> Using the same `AWSClient` across the application and all request threads might cause issues since it's a class, so it will be the same instance across all. The services should get a copy since they're `struct`s but the 'Vapor way' for other stuff is to create a new instance for each request and the application

`AWSClient` holds `AsyncHTTPClient.HTTPClient` and a `CredentialProvider` which both need a shutdown. Passing these around in a struct means there isn't any object that own these objects. I would see `AWSClient` being similar to `Vapor.Application` and `AsyncHTTPClient.HTTPClient` which are both classes.

documentation/DynamoDB and Codable.md
```diff
@@ -40,12 +48,16 @@ let input = DynamoDB.PutItemInput(
...
+
 can be reduced to 
+
 ```swift
 let input = DynamoDB.PutItemCodableInput(item: person, tableName: "my-table")
 let output = try dynamoDB.putItem(input).wait()
 ```
-Similarly when using `DynamoDB.GetItem` the code below which parses an `AttributeValue` dictionary returned from `GetItem`
+
+Similarly when using `DynamoDB.GetItemInput(key:tableName:)`, the code below which parses an `AttributeValue` dictionary returned
```
adam-fowler, line 59: DynamoDB.getItem is the function. DynamoDB.GetItemInput is the input type to that function. I would suggest we use the function name.

documentation/S3 Multipart Upload.md
```diff
@@ -1,8 +1,8 @@
...
+S3 has a series of multipart upload operations. These can be used to upload an object to S3 in multiple parts. If your object is larger than 5GB you are required to use the multipart operations for uploading, but multipart also has the advantage that if one part fails to upload you don't need to re-upload the whole object, just the parts that failed. Finally you can upload multiple parts at the same time and thus improve your upload speed. The multiple part uploading isn't support by Soto at the moment, but it is on the list of improvements to make. You can read more about multipart upload [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html).
```
adam-fowler, line 3: Maybe we should add "You can read Amazon's documentation on multipart upload here"

adam-fowler (commented):
These all look good, just a couple of minor changes

documentation/S3 Multipart Upload.md
0xTim: Do you mean something like

```suggestion
S3 has a series of multipart upload operations. These can be used to upload an object to S3 in multiple parts. If your object is larger than 5GB you are required to use the multipart operations for uploading, but multipart also has the advantage that if one part fails to upload you don't need to re-upload the whole object, just the parts that failed. Finally you can upload multiple parts at the same time and thus improve your upload speed. The multiple part uploading isn't support by Soto at the moment, but it is on the list of improvements to make. You can read  Amazon's documentation on multipart upload [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html).
```

adam-fowler: yep

0xTim:
Requested changes made

adam-fowler (approved):
I think we're good here
thanks

PR #408, opened 2020-10-20
Make DynamoDB.TransactWriteItem an enum

adam-fowler:
Also moved the AttributeValue equatable conformance to the DynamoDB extension code as equatable conformance is not a given for enums so I've stopped generating this automatically.


fabianfett (approved):
This is awesome! Thanks so much!

PR #414, opened 2020-10-28
create a lambda function to test invoke()

sebsto:
As [discussed](https://soto-project.slack.com/archives/C012XA3DE1F/p1603823562069100), the existing LambdaTests assumes a `Hello` Lambda function exists in the testing AWS Account.  This change ensures a Lambda function is created specifically before to run the test, and deletes the function afterwards.

It's a bit longer than I though because IAM Execution Roles are mandatory for Lambda functions, so the code also creates and deletes an role with no permission, to attach to the Lambda function at runtime.

On my Github / AWS accounts, the CI GitHub Actions now runs with success !

### DIFF
--- a/Tests/SotoTests/Services/Lambda/LambdaTests.swift
+++ b/Tests/SotoTests/Services/Lambda/LambdaTests.swift
@@ -12,23 +12,11 @@
 //
 //===----------------------------------------------------------------------===//
 
-/*
-
- The testing code for the AWS Lambda function is created with the following commands:
-
- echo "exports.handler = async (event) => { return \"hello world\" };" > lambda.js
- zip lambda.zip lambda.js
- cat lambda.zip | base64
- rm lambda.zip
- rm lambda.js
-
- */
-
 import NIO
 import XCTest
 
-@testable import SotoIAM
-@testable import SotoLambda
+import SotoIAM
+import SotoLambda
 
 // testing query service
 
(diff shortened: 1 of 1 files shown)


Tests/SotoTests/Services/Lambda/LambdaTests.swift
```diff
@@ -29,15 +113,42 @@ class LambdaTests: XCTestCase {
...
+        do {
+            let response = try Self.createIAMRole().wait()
+            // IAM needs some time after Role creation, before the role can be attached to a Lambda function
+            // https://stackoverflow.com/a/37438525/663360
+            print("Sleeping 15 secs, waiting for IAM Role to be created")
+            sleep(15)
+            _ = try Self.createLambdaFunction(roleArn: response.role.arn).wait()
+        } catch {
+            XCTFail("Can not create prerequisites resources in the cloud, before to execute the tests: \(error)")
+        }
```
adam-fowler, line 141: Can we use `XCTAssertNoThrow` here instead. If you chain the two commands together you don't need to worry about accessing the response from the createIAMRole call.
```
        let response  = Self.createIAMRole().flatMap { response -> EventLoopFuture<Lambda.FunctionConfiguration> in
            let eventLoop = Self.client.eventLoopGroup.next()
            print("Sleeping 15 secs, waiting for IAM Role to be created")
            let scheduled = eventLoop.flatScheduleTask(in: .seconds(15)) {
                Self.createLambdaFunction(roleArn: response.role.arn)
            }
            return scheduled.futureResult
        }
        XCTAssertNoThrow(try response.wait())
```

Tests/SotoTests/Services/Lambda/LambdaTests.swift
```diff
@@ -29,15 +113,42 @@ class LambdaTests: XCTestCase {
...
     }
 
     override class func tearDown() {
+        do {
+            try Self.deleteIAMRole().wait()
+            try Self.deleteLambdaFunction().wait()
+        } catch {
```
adam-fowler, line 148: Again can we use `XCTAssertNoThrow` here

Tests/SotoTests/Services/Lambda/LambdaTests.swift
```diff
@@ -12,15 +12,99 @@
...
+ rm lambda.js
+
+ */
+
 import NIO
-import SotoLambda
 import XCTest
 
+@testable import SotoIAM
+@testable import SotoLambda
```
adam-fowler, line 31: As we aren't access any symbols internal to these frameworks we don't need to import them are @testable

Tests/SotoTests/Services/Lambda/LambdaTests.swift
```diff
@@ -12,15 +12,99 @@
...
+ The testing code for the AWS Lambda function is created with the following commands:
+
+ echo "exports.handler = async (event) => { return \"hello world\" };" > lambda.js
+ zip lambda.zip lambda.js
+ cat lambda.zip | base64
+ rm lambda.zip
+ rm lambda.js
+
+ */
+
```
adam-fowler, line 26: Can we move this comment to just above `createLambdaFunction`

adam-fowler (changes requested):
Looks good, a few comments

Tests/SotoTests/Services/Lambda/LambdaTests.swift
sebsto: nice implementation, also better aligned with other tests

sebsto:
Test fails on APIGateway issue that the other PR should solve. I will trigger a new set of checks once https://github.com/soto-project/soto/pull/417 will be merged.

adam-fowler (commented):
One last thing can we improve `testInvoke` by verifying the contents of the response payload. Aside from that everything is good.

Tests/SotoTests/Services/Lambda/LambdaTests.swift
```diff
@@ -155,10 +155,21 @@ class LambdaTests: XCTestCase {
...
+                // is there an function execution error returned by the service?
+                XCTAssertNil(response.functionError)
+
+                // check the payload matches the one from the Lambda function
+                guard let payload = response.payload?.asString() else {
+                    XCTFail("Lambda function should return a payload")
+                    return
+                }
+                XCTAssertEqual(payload, "\"hello world\"")
+            }
```
adam-fowler, line 172: You've removed the `XCTAssertNoThrow(try response.wait())` so the test doesnt wait for the `invoke` to finish

sebsto (commented):
test suits runs as expected

Tests/SotoTests/Services/Lambda/LambdaTests.swift
adam-fowler: Looks good

adam-fowler (approved):

PR #426, opened 2020-11-03
Reorder public function parameters Logger and EventLoop

adam-fowler:
Place Logger before EventLoop. Will make moving to BaggageContext easier


adam-fowler:
Remember to use branch main before merging

PR #429, opened 2020-11-05
Update Using Soto with Vapor part of README

adam-fowler:


README.md
```diff
@@ -168,14 +168,12 @@ final class MyController {
...
-                let message = SES.Message(body:SES.Body(text:SES.Content(data:emailData.message)), subject:SES.Content(data:emailData.subject))
-                let sendEmailRequest = SES.SendEmailRequest(destination: destination, message: message, source:"awssdkswift@me.com")
+        let emailData = try req.content.decode(EmailData.self)
+        let destination = SES.Destination(toAddresses: [emailData.address])
+        let message = SES.Message(body:SES.Body(text:SES.Content(data:emailData.message)), subject:SES.Content(data:emailData.subject))
+        let sendEmailRequest = SES.SendEmailRequest(destination: destination, message: message, source:"awssdkswift@me.com")
 
-                return client.sendEmail(sendEmailRequest)
-            }
+        return client.sendEmail(sendEmailRequest)
```
0xTim, line 176: would it be better to use `req.client` here to avoid the hop?

README.md
```diff
@@ -168,14 +168,12 @@ final class MyController {
...
             .hopTo(eventLoop: req.eventLoop)
             .map { response -> HTTPResponseStatus in
```
0xTim, line 178: Also, `.transform(to: .ok)` is more 'Vapor-y'

README.md
adam-fowler: You are thinking about soto v5 here surely

README.md
adam-fowler: done

README.md
0xTim: Ah ok I see. Yeah ok leave it as is. I guess the Vapor way would be to do `req.ses.sendEmail` but fitting that into the docs is probably out of scope

adam-fowler: Yeah the V5 docs cover this in a lot more detail, but for V4 I'll stick to this

0xTim (approved):
LGTM

PR #451, opened 2020-12-20
Added CredentialProviderFactory.stsWebIdentityTokenFile 

adam-fowler:
Used in EKS clusters. Looks for environment variables `AWS_WEB_IDENTITY_TOKEN_FILE`, `AWS_ROLE_ARN`, and `AWS_ROLE_SESSION_NAME` to fill out AssumeRoleWithWebIdentity request

### DIFF
--- a/Sources/Soto/Extensions/STS/STS+CredentialProvider.swift
+++ b/Sources/Soto/Extensions/STS/STS+CredentialProvider.swift
@@ -171,16 +171,19 @@ extension STS {
             )
         }
 
+        /// Shutdown credential provider. Need to shutdown webIdentity credential provider
         func shutdown(on eventLoop: EventLoop) -> EventLoopFuture<Void> {
             return tokenPromise.futureResult.flatMap { _ in
-                webIdentityProvider.shutdown(on: eventLoop)
+                self.webIdentityProvider.shutdown(on: eventLoop)
             }
         }
 
+        /// get credentials
         func getCredential(on eventLoop: EventLoop, logger: Logger) -> EventLoopFuture<Credential> {
             return webIdentityProvider.getCredential(on: eventLoop, logger: logger)
         }
 
+        /// Load web identity token file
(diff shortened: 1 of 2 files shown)


Sources/Soto/Extensions/STS/STS+CredentialProvider.swift
```diff
@@ -135,6 +141,83 @@ extension STS {
...
+
+            let tokenPromise = context.eventLoop.makePromise(of: STS.AssumeRoleWithWebIdentityRequest.self)
+            _ = Self.loadTokenFile(tokenFile, on: context.eventLoop).map { token in
+                let request = STS.AssumeRoleWithWebIdentityRequest(
+                    roleArn: roleArn,
+                    roleSessionName: sessionName ?? UUID().uuidString,
+                    webIdentityToken: token
+                )
+                tokenPromise.succeed(request)
+            }
```
t089, line 163: I think you would need to load the token file every time you want to make a new `AssumeRoleWithWebIdentityRequest`. So the `loadTokenFile` call should be part of the `requestProvider`. The reason is that the token itself might expire. EKS will automatically update the projected file with a new token though. So as long as you always read the token from disk, you should always be fine.

t089 (changes requested):
I can confirm that this work on an AWS EKS cluster.

```swift
import SotoSTS

let client = AWSClient(credentialProvider: .stsWebIdentityTokenFile(region: .euwest1),
                       httpClientProvider: .createNew)
let sts = STS(client: client)
let whoami = try sts.getCallerIdentity(.init()).wait()
print(whoami)
```

I think the request provider would need to be changed to read the token file every time it tries to call the `AssumeRoleWithWebIdentity` endpoint.

Sources/Soto/Extensions/STS/STS+CredentialProvider.swift
adam-fowler: ok will fix

adam-fowler:
Do you want to give the latest a go

t089 (approved):
LGTM, thanks a lot! 🚀

Issue #471, opened 2021-02-09
Aurora Serverless Cluster ARNs can have more than 100 characters.


DwayneCoussement:
**Describe the bug**
Aurora Serverless Cluster ARNs can have more than 100 characters.

**Suggestion**
Change the max length to 2048 https://docs.aws.amazon.com/IAM/latest/APIReference/API_Policy.html

**To Reproduce**
1. Create a new aurora serverless cluster (via CDK for example)

```    
    const vpc = new ec2.Vpc(this, 'SotoVPC');
    const dbName = 'SotoDB';
    const cluster = new rds.ServerlessCluster(this, 'SotoCluster', {
      engine: rds.DatabaseClusterEngine.AURORA_POSTGRESQL,
      parameterGroup: rds.ParameterGroup.fromParameterGroupName(this, 'ParameterGroup', 'default.aurora-postgresql10'),
      defaultDatabaseName: dbName,
      vpc
    });
```
2. fetch clusterArn via `cluster.clusterArn`
3. Make a request:
```
let query = RDSDataService.ExecuteStatementRequest(database: "SotoDB", resourceArn: clusterArn, secretArn: secretArn)
self.rds.executeStatement(query).wait()
```
4. Notice you get a validation error:
ValidationError: Length of ExecuteStatementRequest.resourceArn (101) is greater than the maximum allowed value 100.

**Expected behavior**
Don't make an assumption on that the ARN will be less than 100

**Setup (please complete the following information):**
 - OS: Amazon Linux2
 - Version of soto: 5.2
 - Authentication mechanism [hard-coded credentials, IAM Instance Profile on EC2, etc]

**Additional context**
Add any other context about the problem here. If this a bug related to a command not working can you run the same command on a client initialised with the logging middleware. eg `S3(region: .useast1, middlewares:[AWSLoggingMiddleware()])`

adam-fowler:
#473

PR #472, opened 2021-02-09
Increase ARN size for RDS to max allowed

DwayneCoussement:
See #471

### DIFF
--- a/Sources/Soto/Services/RDSDataService/RDSDataService_Shapes.swift
+++ b/Sources/Soto/Services/RDSDataService/RDSDataService_Shapes.swift
@@ -94,11 +94,11 @@ extension RDSDataService {
         public func validate(name: String) throws {
             try self.validate(self.database, name: "database", parent: name, max: 64)
             try self.validate(self.database, name: "database", parent: name, min: 0)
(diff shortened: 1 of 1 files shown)


adam-fowler:
The limit on ARN size is coming from the AWS model files. The code you are editing is generated from these files. The next time we re-generate code your changes will be lost. The CodeGenerator has a method for patching the JSON models before they are used to generate the swift files. 

You need to edit `[patch.swift](https://github.com/soto-project/soto/blob/main/CodeGenerator/Sources/CodeGenerator/patch.swift)` in the CodeGenerator for your changes to persist. It's a little awkward to get right as you have to look through the model file to find out what you need to edit and then the patch uses an array of KeyPaths to get a `WriteableKeyPath` for the part of the JSON model to edit. I've added a separate PR to do this #473

Interestingly the 100 characters limit is in the documentation as well https://docs.aws.amazon.com/rdsdataservice/latest/APIReference/API_ExecuteStatement.html

DwayneCoussement:
Hey @adam-fowler, thank you for your feedback and follow up. I've also provided this feedback to Amazon on the mentioned page. If I hear anything back, I'll drop a comment somewhere :-)

PR #479, opened 2021-02-18
Add EXTERNAL_PROVIDER to CognitoIdentityProvider.UserStatusType

adam-fowler:


t089 (approved):
Nice

t089:
Hm I m a bit worried because this does not seem to exist in [the Java SDK](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-cognitoidp/src/main/java/com/amazonaws/services/cognitoidp/model/UserStatusType.java)

t089:
Ah, note though that while the enum case does not exist in the sdk, the enum itself is actually not used to represent the status of a user.

https://github.com/aws/aws-sdk-java/blob/c1e356268d07d8fa8f524a74394b0d6702d79808/aws-java-sdk-cognitoidp/src/main/java/com/amazonaws/services/cognitoidp/model/UserType.java#L532

adam-fowler:
The Java SDK doesn't have it either as they build their code off the same model files I use. It is a bug in the model file and this needs to be passed back to the cognito team.

PR #488, opened 2021-04-12
Async/await

adam-fowler:
Related SotoCore PR is [here](https://github.com/soto-project/soto-core/pull/431)

Async/await additions
- Add async versions of all standard API calls in `*_API+async` files
- Add AsyncSequence paginators in `*_Paginator+async` files
- Add async versions of waiters in `*_Waiter+async` files
- Replicate S3 and TimeStreamWrite tests using async apis

Last tested with swift 5.5 snapshot from 15 Sep 2021 of Swift compiler


Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable+async.swift
```diff
@@ -0,0 +1,150 @@
...
+    ///
+    /// Invalid Requests with empty values will be rejected with a `ValidationException` exception.</p> <note> <p>To prevent a new item from replacing an existing item, use a conditional expression that contains the `attribute_not_exists` function with the name of the attribute being used as the partition key for the table. Since every record must contain that attribute, the `attribute_not_exists` function will only succeed if no matching item exists.</p> </note> <p>For more information about `PutItem`, see <a href="https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html">Working with Items</a> in the <i>Amazon DynamoDB Developer Guide</i>.
+    public func putItem<T: Encodable>(
+        _ input: PutItemCodableInput<T>,
+        logger: Logger = AWSClient.loggingDisabled,
+        on eventLoop: EventLoop? = nil
+    ) async throws -> PutItemOutput {
+        try await self.putItem(input, logger: logger, on: eventLoop).get()
```
0xTim, line 35: We should use the `async` version of `putItem` instead of the ELF version to make the transition easier longer term IMO. Same goes for everything in this file

0xTim (changes requested):
IMO we should avoid the use of `.get()` in Soto and leave it to SotoCore calling get on the client execute function. Will make future work easier

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable+async.swift
adam-fowler: Fair point. Will fix. I should probably do something similar for S3.multipart but that is a fair amount of work so will do in another PR

adam-fowler: Just had a look at this and this would mean replicating the code in DynamoDB+Codable is that what you intended?

adam-fowler: Added them in so you can see changes

0xTim: Ah right, no that's not quite what I intended actually. I didn't realise it was going to involve that. However, I definitely think it's the longer term solution, but agree that it's probably overkill for each service addition until we actually need it

0xTim: (Other option of course is to extract all the code that's duplicated to avoid duplicating all the logic)

adam-fowler: > (Other option of course is to extract all the code that's duplicated to avoid duplicating all the logic)

Did as you said.
Also added some tests to ensure it all works

0xTim (approved):
LGTM!

PR #498, opened 2021-05-25
Add support for Waiters

adam-fowler:
Waiters are a client-side abstraction used to poll a resource until a desired state is reached, or until it is determined that the resource will never enter into the desired state. This is a common task when working with services that are eventually consistent like Amazon S3 or services that asynchronously create resources like Amazon EC2. Writing logic to continuously poll the status of a resource can be cumbersome and error-prone. The goal of waiters is to move this responsibility out of customer code and onto Soto.

See https://github.com/soto-project/soto-core/pull/442 also


adam-fowler:
This PR has now been replaced by #504

PR #504, opened 2021-07-09
Add support for waiter using JMESPath

adam-fowler:
Waiters are a client-side abstraction used to poll a resource until a desired state is reached, or until it is determined that the resource will never enter into the desired state. This is a common task when working with services that are eventually consistent like Amazon S3 or services that asynchronously create resources like Amazon EC2. Writing logic to continuously poll the status of a resource can be cumbersome and error-prone. The goal of waiters is to move this responsibility out of customer code and onto Soto.

See https://github.com/soto-project/soto-core/pull/446
This implementation uses JMESPath for path matching.

A simple example of usage would be creating a DynamoDB table and waiting until it is ready to be used
```swift
let input = DynamoDB.CreateTableInput(
    ...
    tableName: name
)
return dynamoDB.createTable(input).flatMap { _ in
    dynamoDB.waitUntilTableExists(.init(tableName: name))
}
```


CodeGenerator/Sources/CodeGenerator/Models/Waiter.swift
```diff
@@ -0,0 +1,106 @@
...
+// See CONTRIBUTORS.txt for the list of Soto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+// used to decode waiters-2.json files
+struct Waiters: Decodable {
+    struct Waiter: Decodable {
+        enum State: String, Decodable {
```
0xTim, line 18: Does it make sense to pull this from core rather than redefining it? (Relates to the question about whether the enum will ever change)

CodeGenerator/Sources/CodeGenerator/Patch/patchTypes.swift
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    override func apply(to api: inout Root) throws {
+        guard let _ = api[patchKeyPath: patchKeyPath] else { throw APIPatchError.doesNotExist }
+        api[patchKeyPath: self.patchKeyPath]?[self.key] = self.value
+    }
+}
+
+extension Shape.ShapeType: Equatable {
+    /// use to verify is shape types are the same when checking original values in replace patches
```
0xTim, line 112:
```suggestion
    /// use to verify if shape types are the same when checking original values in replace patches
```

Sources/Soto/Services/EC2/EC2_Shapes.swift
```diff
@@ -336,6 +336,7 @@ extension EC2 {
         case cancelled
         case cancelling
         case completed
+        case deleted = "DELETED"
```
0xTim, line 339: Why is this the only one that's uppercased?

0xTim (approved):
LGTM - few nits

CodeGenerator/Sources/CodeGenerator/AWSService+waiters.swift
```diff
@@ -0,0 +1,156 @@
...
+        let matcher: MatcherContext
+    }
+
+    enum MatcherContext {
+        case jmesPath(path: String, expected: String)
+        case jmesAnyPath(path: String, expected: String)
+        case jmesAllPath(path: String, expected: String)
+        case error(String)
+        case errorStatus(Int)
+        case success(Int) // success requires a associated value, so a mustache context is created for the value
```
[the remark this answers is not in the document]
adam-fowler: Its just a dummy value, a mustache context won't be created for `success` if there is no associated value. New comment
"// success requires a dummy associated value, so a mustache context is created for the `MatcherContext`"

CodeGenerator/Sources/CodeGenerator/CodeGenerator.swift
```diff
@@ -73,7 +73,21 @@ struct CodeGenerator {
...
+            var waiters: Waiters?
+            do {
+                if let waiterFile = Glob.entries(pattern: $0 + "/**/waiters-*.json").first {
+                    let waiterData = try Data(contentsOf: URL(string: "file://\(waiterFile)")!)
+                    waiters = try JSONDecoder().decode(Waiters.self, from: waiterData)
+                    try waiters?.patch(serviceName: api.serviceName)
+                } else {
+                    waiters = nil
+                }
+            } catch {
```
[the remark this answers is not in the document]
adam-fowler: debug code, removed

CodeGenerator/Sources/CodeGenerator/Models/Waiter.swift
adam-fowler: The CodeGenerator is not dependent on soto-core

Sources/Soto/Services/EC2/EC2_Shapes.swift
adam-fowler: Looks like my patch was wrong. The EC2 models don't have the `deleted` state in them, but the EC2 waiter models depend on this state being there. I patched the EC2 models, but looking at the waiter I did it wrong.

PR #507, opened 2021-07-19
Add support for Host prefix values

adam-fowler:
A number of API calls have a prefix they prepend to the host name (chime, location, s3control etc). This PR adds the hostPrefix parameter to all the AWSClient.execute functions where required. 

See related PR https://github.com/soto-project/soto-core/pull/448


0xTim (approved):
LGTM

PR #513, opened 2021-08-17
JSON extension for RDSDataService.ExecuteStatementResponse

patagoniacode:
### Description
When using Soto for interacting with the Aurora Serverless V1 Data API, it would be useful to receive the database results as a JSON object or JSON data. This could be also helpful for decoding the response to a Swift object.

Currently, the return type of the expression RDSDataService.executeStatement(statement) is of type RDSDataService.ExecuteStatementResponse.

The current initial implementation adds two methods to the RDSDataService.ExecuteStatementResponse object:
- jsonString() throws -> String // This function can be called on self and returns a JSON string with the database results.
- jsonData() throws -> Data // This function can be called on self and returns JSON data with the database results.

### Example usage

```
import Foundation
import SotoRDSDataService

// Swift object
struct Product: Codable, Hashable {
    let product_id: Int
    let name: String
}

// Aurora Serverless:
     // Table 'product'
          // The table has two columns:
               // 'product_id': INT(255)
               // 'name': VARCHAR(255)

struct AuroraManager {
    static func executeStatement(sqlRequest: String, clusterArn: String, databaseName: String, secretArn: String, accessKeyId: String = AWSAccessKeys.accessKeyId, secretAccessKey: String = AWSAccessKeys.secretAccessKey) throws -> RDSDataService.ExecuteStatementResponse {
        let client = AWSClient(credentialProvider: .static(accessKeyId: accessKeyId, secretAccessKey: secretAccessKey), httpClientProvider: .createNew)
        defer { try? client.syncShutdown() }
        let rds = RDSDataService(client: client)
        let statement = RDSDataService.ExecuteStatementRequest(continueAfterTimeout: true, database: databaseName, includeResultMetadata: true, parameters: nil, resourceArn: clusterArn, secretArn: secretArn, sql: sqlRequest)
        let response = try rds.executeStatement(statement).wait()
        return response
    }
}

```

The call site would be like this:
```
do {
     let response = try AuroraManager.executeStatement(sqlRequest: sqlRequest, clusterArn: clusterArn, databaseName: databaseName, secretArn: secretArn)
     let jsonData = try response.jsonData()
     let jsonDecoder = JSONDecoder()
     let products = try jsonDecoder.decode([Product].self, from: jsonData)
     } catch {
     print(error)
}
```

This is my first PR, so any guidance would be highly appreciated.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -505,7 +505,8 @@ let package = Package(
         .target(name: "SotoQuickSight", dependencies: [.product(name: "SotoCore", package: "soto-core")], path: "./Sources/Soto/Services/QuickSight"),
         .target(name: "SotoRAM", dependencies: [.product(name: "SotoCore", package: "soto-core")], path: "./Sources/Soto/Services/RAM"),
         .target(name: "SotoRDS", dependencies: [.product(name: "SotoCore", package: "soto-core")], path: "./Sources/Soto/Services/RDS"),
-        .target(name: "SotoRDSDataService", dependencies: [.product(name: "SotoCore", package: "soto-core")], path: "./Sources/Soto/Services/RDSDataService"),
+        .target(name: "SotoRDSDataService", dependencies: [.product(name: "SotoCore", package: "soto-core")], path: "./Sources/Soto/", sources: ["Services/RDSDataService", "Extensions/RDSDataService"]),
+        
         .target(name: "SotoRedshift", dependencies: [.product(name: "SotoCore", package: "soto-core")], path: "./Sources/Soto/Services/Redshift"),
         .target(name: "SotoRedshiftDataAPIService", dependencies: [.product(name: "SotoCore", package: "soto-core")], path: "./Sources/Soto/Services/RedshiftDataAPIService"),
         .target(name: "SotoRekognition", dependencies: [.product(name: "SotoCore", package: "soto-core")], path: "./Sources/Soto/Services/Rekognition"),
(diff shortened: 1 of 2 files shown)


adam-fowler:
While this does work for most cases this is probably not the best way to do this. You are basically decoding the data twice: once to convert to JSON and then when you run `JSONDecoder`. It would be much more preferable to have a `Decoder` that goes straight from the `ExecuteStatementResponse` to your `Codable` type.

patagoniacode:
Hi @adam-fowler, thank you for your feedback.
The modified implementation adds the following method to the RDSDataService.ExecuteStatementResponse object:

` public func decode<T: Decodable>(_ type: T.Type) throws -> T`

It allows the call site to be like this:

```
do {
     let response = try AuroraManager.executeStatement(sqlRequest: sqlRequest, clusterArn: clusterArn, databaseName: databaseName, secretArn: secretArn)
     let products = try response.decode([Product].self)
     } catch {
     print(error)
}
```

adam-fowler:
I don't think you understood my comment. I was saying I don't think it is a good idea to use JSONDecoder() as you are converting the data through multiple formats before getting you result ie `ExecuteStatementResponse` -> JSON -> Codable type. What you want is `ExecuteStatementResponse` -> Codable type. 

Checkout https://github.com/soto-project/soto/Sources/Soto/Extensions/DynamoDB/DynamoDBDecoder.swift.

adam-fowler:
Closing as this did not move forward

PR #515, opened 2021-08-20
Fix completedParts array in resumeMultipartUpload error

adam-fowler:
Return new `ResumeMultipartUploadRequest` with updated `completedParts` array when `resumeMultipartUpload` fails. Previously we were just using the original resume request and the `completedParts` never got updated.


0xTim (approved):
LGTM

PR #517, opened 2021-08-25
Endpoint discovery for DynamoDB and TimeStream

adam-fowler:
This PR adds endpoint discovery for services DynamoDB and TimeStream. 

TimeStream will run the endpoint discovery and api call one after the other as the api call requires the endpoint returned by the endpoint discovery. DynamoDB runs the endpoint discovery and api call in parallel because it can use it's original endpoint while the endpoint discovery has not returned an endpoint.

By default endpoint discovery is disabled for DynamoDB because it is not needed but if you want to enable it there is a service config option to enable it.
```swift
let dynamoDB = DynamoDB(client: awsClient, options: .enableEndpointDiscovery)
```

Related SotoCore PR is https://github.com/soto-project/soto-core/pull/457


CodeGenerator/mustache/api.mustache
```diff
@@ -110,10 +117,20 @@ public struct {{ name }}: AWSService {
...
+        return self.client.execute(operation: "{{name}}", path: "{{path}}", httpMethod: .{{httpMethod}}, serviceConfig: self.config{{#inputShape}}, input: input{{/inputShape}}{{#endpointDiscovery}}, endpointDiscovery: .init(storage: self.endpointStorage, discover: self.getEndpoint, required: {{required}}){{/endpointDiscovery}}{{#hostPrefix}}, hostPrefix: "{{{.}}}"{{/hostPrefix}}, logger: logger, on: eventLoop{{#streaming}}, stream: stream{{/streaming}})
     }
 {{/streamingOperations}}
 {{/first(streamingOperations)}}
+{{#endpointOperation}}
+
+    func getEndpoint(logger: Logger, eventLoop: EventLoop) -> EventLoopFuture<AWSEndpoints> {
+        return {{.}}(.init(), logger: logger, on: eventLoop).map {
+            .init(endpoints: $0.endpoints.map {
+                .init(address: "https://\($0.address)", cachePeriodInMinutes: $0.cachePeriodInMinutes)
```
0xTim, line 129: Can endpoints be HTTP for endpoint discovery? I know that locally I use HTTP for DynamoDB but that seems like a different use case to the endpoint discovery described in the docs after a skim read

0xTim (approved):
LGTM. One query and obviously need to update the manifest when the Soto Core is merged

CodeGenerator/mustache/api.mustache
adam-fowler: Do you mean HTTP instead of HTTPS? We are only ever provided an address without a scheme. So the assumption is it would be HTTPS.

Endpoints and DynamoDB are a little confused. They have VPC endpoints, but that is a different thing from what I can tell. When using the endpoint discovery I haven't managed to get DynamoDB to return an endpoint different from the actual one I originally used. I am assuming you need much larger table usage to get multiple endpoints. This is one of the reasons DynamoDB endpoint discovery is disabled by default apart from the fact it isn't required.

Would love to see if someone with a much larger dataset to work with could get different results. You just need to run `DynamoDB.DescribeEndpoints` to see if you'd get a different endpoint. If you do you can use a `DynamoDB` object with the `.enableEndpointDiscovery` flag set.

0xTim: I've tried with a client project (millions of items across 10s of tables) and it's the default endpoints. The table size shouldn't matter because the router should always be the same, it's just a matter of whether your data is held in one partition or 100s of partitions, but AFAIK that should be transparent to the client.

I'm guessing it's just them future proofing themselves and is unrelated to an endpoint used for local testing so I'm happy

PR #519, opened 2021-08-27
resumeMultipartUpload recursive function call fix

adam-fowler:
If you call `whenComplete`, `map` etc on a completed `EventLoopFuture` if will call the closure immediately. This was causing `resumeMultipartUpload` to overflow the stack if you were uploading a large enough object and enough of the parts could be skipped.

We avoid the stack overflow by putting the recursive call inside a `EventLoop.submit` call.


0xTim (approved):
LGTM

PR #526, opened 2021-09-17
Replace NIO with NIOCore where applicable

adam-fowler:


0xTim (approved):
LGTM. The only thing I don't see is adding `NIOCore` as a dependency in the manifest - was this done previously or are you relying on SwiftPM exposing that to you via NIO's exports?

adam-fowler:
> LGTM. The only thing I don't see is adding `NIOCore` as a dependency in the manifest - was this done previously or are you relying on SwiftPM exposing that to you via NIO's exports?

I guess the services that include the `import NIOCore/NIOPosix` should really have the dependency in the manifest. It's a bit of a pain, when you are autogenerating the `Package.swift`. In most cases I could actually remove the `import NIOCore` as I have `@_exported` a number of the `NIOCore` symbols from `SotoCore`.

0xTim:
Cool as long as it's something we export that should be fine. My thinking was to avoid headaches in the future when SwiftPM fixes the bug of making all targets in the dependency tree available to everything at some point in the future

adam-fowler:
> Cool as long as it's something we export that should be fine. My thinking was to avoid headaches in the future when SwiftPM fixes the bug of making all targets in the dependency tree available to everything at some point in the future

I'll look into this in another PR maybe

PR #529, opened 2021-09-22
Use mustache for generating Package.swift

adam-fowler:
Removing anothe part of the project dependent on Stencil


0xTim (approved):
I'm not familiar with Mustache syntax but the changes look sensible and I'm sure we'd find out pretty quick if anything was broken!

PR #534, opened 2021-10-09
Add CognitoIdentity renewing external identity provider

adam-fowler:
Previously there was no way to renew the credentials that generated the Identity required for the Cognito Identity credential provider. So when the identity credentials expired you had to create a new `AWSClient`. This PR adds a new `IdentityProvider` via `.externalIdentityProvider` which renews the identity as well when credentials expire. The closure included is called every time a new identity is needed. It returns a set of name-value pairs that map provider names to provider tokens.

Here is an example using Cognito Userpools for identity. In this case the closure is returning 
  
```swift
let credentialProvider: CredentialProviderFactory = .cognitoIdentity(
    identityPoolId: "my-identiy-pool-id",
    identityProvider: .externalIdentityProvider(tokenProvider: { context in
        let userPoolIdentityProvider = "cognito-idp.\(context.region).amazonaws.com/\(userPoolId)"
        let cognitoIdentityProvider = CognitoIdentityProvider(client: context.client, region: context.region)
        let request = CognitoIdentityProvider.InitiateAuthRequest(
            authFlow: .userPasswordAuth,
            authParameters: ["USERNAME": "my-username", "PASSWORD": "my-password"],
            clientId: "my-client-id"
        )
        return cognitoIdentityProvider.initiateAuth(request, logger: context.logger, on: context.eventLoop)
            .flatMapThrowing { response in
                guard let idToken = response.authenticationResult?.idToken else { throw CredentialProviderError.noProvider }
                return [userPoolIdentityProvider: idToken]
            }
    }),
    region: .euwest1
)
```

### DIFF
--- a/Sources/Soto/Extensions/CognitoIdentity/CognitoIdentity+IdentityProvider.swift
+++ b/Sources/Soto/Extensions/CognitoIdentity/CognitoIdentity+IdentityProvider.swift
@@ -127,10 +127,10 @@ extension CognitoIdentity {
                 eventLoop: eventLoop,
                 logger: logger
             )
-            return loginsProvider(context)
+            return self.loginsProvider(context)
                 .flatMap { logins in
                     let request = CognitoIdentity.GetIdInput(identityPoolId: context.identityPoolId, logins: logins)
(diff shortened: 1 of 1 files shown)


adam-fowler:
@0xTim using this to implement a full Cognito user pool implementation here https://github.com/soto-project/soto-cognito-authentication-kit/pull/22

PR #540, opened 2021-10-19
v6: Remove old code generator now that we are using soto-codegenerator

adam-fowler:
https://github.com/soto-project/soto-codegenerator


adam-fowler:
Errors from this PR are fixed in #539

PR #547, opened 2021-11-10
Add Expect: 100-continue header for large S3 uploads

adam-fowler:
Means requests will error immediately instead of waiting until upload has completed

Requires AHC v1.6.4, so I need to update SotoCore as well


adam-fowler:
Closing as we should be merging into 6.x.x

PR #551, opened 2021-11-16
Async versions of Multipart Upload/Download/Copy

adam-fowler:
Rewrote them completely to use swift concurrency


adam-fowler:
This branch requires 6.x.x to be merged

Sources/Soto/Extensions/S3/S3+multipart.swift
```diff
@@ -34,6 +34,24 @@ extension S3 {
...
+                threadPool.start()
+                return threadPool
+            case .shared(let sharedPool):
+                return sharedPool
+            }
+        }
+
+        func destory(_ threadPool: NIOThreadPool) {
+            if case .createNew = self {
+                threadPool.shutdownGracefully { _ in }
```
0xTim, line 52: Would it make sense to use `syncShutdownGracefully` here to account for the async nature of it? I'm aware this then throws etc and can ripple everywhere

0xTim (approved):
LGTM

Sources/Soto/Extensions/S3/S3+multipart.swift
adam-fowler: This is used by both async and non-async code. Just noticed a typo though `destroy` not `destory`.
I guess it would be useful if there was an async version of this. I could easily add one.
```swift
extension S3.ThreadPoolProvider {
    func destroy(_ threadPool: NIOThreadPool) async throws {
        if case .createNew = self {
            return try await withUnsafeThrowingContinuation { cont in
                threadPool.shutdownGracefully { error in
                    if let error = error {
                        cont.resume(throwing: error)
                    } else {
                        cont.resume()
                    }
                }
            }
        }
    }
}
```

PR #557, opened 2021-11-23
Deprecated member initializers

### LINKED ISSUE #352: Find way to deprecate AWSShape members
Currently deprecated members are removed. But if we keep them around and add a `@available(*, deprecate, ...) to deprecate them, you get a warning in the init() function where it is initialised. Need a better solution

adam-fowler:
For shapes with deprecated member variables 
- Remove deprecated variables from standard `init` function
- Add an additional `init` which is flagged as deprecated which includes deprecated variables

Soto Code generator PR is https://github.com/soto-project/soto-codegenerator/pull/34
This resolves #352

### DIFF
--- a/Sources/Soto/Services/AppStream/AppStream_Shapes.swift
+++ b/Sources/Soto/Services/AppStream/AppStream_Shapes.swift
@@ -3471,6 +3471,30 @@ extension AppStream {
         /// The VPC configuration for the fleet. This is required for Elastic fleets, but not required for other fleet types. Elastic fleets require that you specify at least two subnets in different availability zones.
         public let vpcConfig: VpcConfig?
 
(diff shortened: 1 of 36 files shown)


0xTim:
Is there a reason to keep a deprecated initialiser around if we're prepping a major release?

adam-fowler:
> Is there a reason to keep a deprecated initialiser around if we're prepping a major release?

These are variables deprecated by AWS. I won't be able to tell the difference between a newly deprecated variable and one deprecated prior to v6, so won't know if I can delete it. So am just sticking to deprecating all of them. AWS's versioning of model files is separate from the Soto version.

0xTim (approved):
Ah fair enough

PR #562, opened 2021-11-24
Fix capitalisation of Swift variables, labels and enums

adam-fowler:
See https://github.com/soto-project/soto-codegenerator/pull/36 for details

This does change quite a number of enum, variable and label names but they make much better sense now.


Yasumoto:
Wow, this is a huge improvement 😲 

Left a few questions on the PR in soto-core, but overall this is phenomenal.

PR #568, opened 2022-01-01
Fix CloudFront patch for both DistributionConfig and DistributionSummary

adam-fowler:


adam-fowler:
Changes were merged in from model update PR

PR #587, opened 2022-04-01
Function does not preserve the trailing / if present

jmoisan21:
virtualAddressFixup does not preserve, or more precisely restore, the trailing / if it was present. That results in everything being considered as a files where the original intent was to define a folder.

### DIFF
--- a/Sources/Soto/Extensions/S3/S3RequestMiddleware.swift
+++ b/Sources/Soto/Extensions/S3/S3RequestMiddleware.swift
@@ -87,6 +87,10 @@ public struct S3RequestMiddleware: AWSServiceMiddleware {
                 urlPath = paths.joined(separator: "/")
                 urlHost = host
             }
+            // add trailing "/" back if it was present
+            if request.url.pathWithSlash.hasSuffix("/") {
+                urlPath += "/"
+            }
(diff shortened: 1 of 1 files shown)


adam-fowler:
I'm happy to merge this but can I ask what the intent is here? What S3 operation do you need this for? 

S3 does not have a file system. It is just a list of keys with associated files. There are no folders. A file system is faked by pretending "/" means something when it doesn't.

jmoisan21:
> What S3 operation do you need this for?

S3.PutObjectRequest is the one that made me discover this issue. Others may also be impacted but I would have to double check if they really are (such as S3.PutObjectAclRequest).

> S3 does not have a file system. It is just a list of keys with associated files. There are no folders. A file system is faked by pretending "/" means something when it doesn't.

Agreed and absolutely true.
In reality if you create multiple files under the same root key (ie /fake/folder/file1 and /fake/folder/file2) then S3 will allow you to go and apply ACL to /fake or /folder. Essentially allowing to simulate/emulate that those keys are directories.
Those false folders keys representations can become useful in creating a hierarchy with rights applied to them before other clients put files with those keys.

adam-fowler (approved):

jmoisan21:
TY

PR #626, opened 2022-10-21
Enforce KMS parameter requirements on data key request object initializers

MPLew-is:
KMS requires exactly one of `keySpec` and `numberOfBytes` for both `GenerateDataKey` and `GenerateDataKeyPair`, but the current request object initializers allow one or both to be `nil`. Right now, getting this wrong would only be caught at runtime, whereas these new initializers would enforce this behavior at compile-time.

While this change will not alter behavior for clients using the `KMS` module directly (the KMS service will reject all requests that don't meet these requirements anyways), removing the original initializers is still a source-level breaking change that might cause build failures for any client code wrapping the Soto KMS calls. Thus, the original initializers have just been marked as deprecated for now, with full removal saved for a major version bump.

This PR is mostly to just start a conversation about the mismatch in KMS parameter requirements and the corresponding Swift initializers, with a starting implementation if it's decided this is the correct path.


adam-fowler:
This file is generated from the Smithy model files AWS provides. Here is the one for KMS. https://github.com/soto-project/soto/blob/main/models/kms.json. I get these files from the aws-sdk-go-v2 repo. You cannot edit the swift service files directly as any edits will be lost the next time the model files are updated and service files are re-generated.

MPLew-is:
Ah, I see - apologies for the confusion. Since there doesn't seem to be a concept of "one of these must be provided" in the model files I guess I'll continue just bandaging this with custom extensions.

PR #628, opened 2022-10-26
Fix default trait processing

adam-fowler:
I was processing the default trait incorrectly. Clients should only use the default trait when the required trait is also present. In any other situation the default trait is a server only trait.

Related Code Generator commit is [461c2fe33a31eb8ee703fc4ed80601f56835c7d3](https://github.com/soto-project/soto-codegenerator/commit/461c2fe33a31eb8ee703fc4ed80601f56835c7d3)


0xTim (approved):
LGTM - have you got a major release planned for this as it makes a load of stuff in the shapes optional that weren't

adam-fowler:
It is reverting changes made earlier that were never released so it isn't an issue

PR #648, opened 2022-12-30
Rebuild service files with latest Code Generator

adam-fowler:
Run the latest code generator on current models. This merges a number of the service files and produces a consistent number of files for each service (api, shapes, api+async) as required by build plugin. 

I changed some capitalisation for file names so built this on Linux, as Linux git copes with capitalisation better then macOS does


0xTim (approved):
Checked a few files and LGTM

PR #671, opened 2023-05-13
Allow more complex ConditionExpressions using UpdateItemCodableInput

Andrea-Scuderi:
# Description

The current implementation of `DynamoDB.UpdateItemCodableInput` doesn't not allow to implement optimistic locking.
In fact, ConditionExpressions in that struct can compare the current value of a single attribute with a known value.
However, many applications require more complex condition expressions that involve multiple attributes and operators. 
For example, consider the following case:

```
struct Item: Codable {
   let key: String
   let name: String
   let createdAt: String
   let updatedAt: String
}
```

with the current implementation we can update the DynamoDB table in this way:
```
let item = Item(
   key: "9D1DC963-411A-486F-BE26-ADF473F00B80",
   name: "MyItem"
   createdAt: "2023-05-08T18:17:34.972Z",
   updatedAt: "2023-05-11T12:18:41.699Z"
)
let item = try DynamoDB.UpdateItemCodableInput(
            conditionExpression: "attribute_exists(#createdAt)",
            key: [keyName],
            tableName: tableName,
            updateItem: item
        ).createUpdateItemInput()
let _ = try await db.updateItem(input)
```

The previous command has a problem though, there is no way to add a condition on the key and, in a concurrent scenario there is no way to prevent an update if updatedAt field doesn't match with the current value on the DynamoDB table.

Having additionalAttributeNames and additionalAttributeValues as parameter of the `DynamoDB.UpdateItemCodableInput` will allow to write more complex `conditionExpression ` without loosing the flexibility introduced by the generated `expressionAttributeValues` and `expressionAttributeNames`.

```
let item = Item(
   key: "9D1DC963-411A-486F-BE26-ADF473F00B80",
   name: "MyItem"
   createdAt: "2023-05-08T18:17:34.972Z",
   updatedAt: "2023-05-12T12:18:41.699Z"
)
let oldUpdatedAt = "2023-05-11T12:18:41.699Z"
let item = try DynamoDB.UpdateItemCodableInput(
            additionalAttributeNames: ["#keyName": keyName],
            additionalAttributeValues: [":oldUpdatedAt" : .s(oldUpdatedAt)],
            conditionExpression: "attribute_exists(#keyName) AND #updatedAt = :oldUpdatedAt",
            key: [keyName],
            tableName: tableName,
            updateItem: item
        )
let _ = try await db.updateItem(input)
```
The previous update will update the table only if the key exists and the `updateAt` value in the table match the value `oldUpdatedAt` of the request, while the new value is set to `Item.updatedAt`.

In the previous example the `additionalAttributeNames` dictionary will be merged to `expressionAttributeNames` and `additionalAttributeValues` to `expressionAttributeValues`.

# Implementation

- The implementation introduces a non breaking change
- Add `additionalAttributeNames` and `additionalAttributeValues` as nullable in the init method of `DynamoDB.ConditionalUpdateItemCodableInput`.
- In case `expressionAttributeNames` or `updateExpression` are `nil` and  `additionalAttributeNames` is not `nil`, the content of `additionalAttributeNames` will be merged to `expressionAttributeNames` overriding existing keys.
- If `additionalAttributeValues` is not `nil`, the content of `additionalAttributeValues` will be merged to `expressionAttributeValues` overriding existing keys.

### DIFF
--- a/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
+++ b/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
@@ -377,7 +377,7 @@ extension DynamoDB {
             let expressionAttributeNames: [String: String]
             if let names = self.expressionAttributeNames, self.updateExpression != nil {
                 expressionAttributeNames = names
-            } else if let additionalAttributeNames {
+            } else if let additionalAttributeNames = additionalAttributeNames {
                 let tmpAttributeNames: [String: String] = .init(item.keys.map { ("#\($0)", $0) }) { first, _ in return first }
                 expressionAttributeNames = tmpAttributeNames.merging(additionalAttributeNames, uniquingKeysWith: { _, new in new })
             } else {
@@ -385,7 +385,7 @@ extension DynamoDB {
             }
 
             let expressionAttributeValues: [String: AttributeValue]
-            if let additionalAttributeValues {
+            if let additionalAttributeValues = additionalAttributeValues {
                 let tmpExpressionAttributeValues: [String: AttributeValue] = .init(item.map { (":\($0.key)", $0.value) }) { first, _ in return first }
                 expressionAttributeValues = tmpExpressionAttributeValues.merging(additionalAttributeValues, uniquingKeysWith: { _, new in new })
             } else {
--- a/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
+++ b/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
@@ -349,9 +349,9 @@ extension DynamoDB {
         /// Object containing item key and attributes to update
         public let updateItem: T
 
-        public init(additionalAttributeNames: [String: String]? = nil, additionalAttributeValues: [String: AttributeValue]? = nil, conditionExpression: String? = nil, expressionAttributeNames: [String: String]? = nil, key: [String], returnConsumedCapacity: ReturnConsumedCapacity? = nil, returnItemCollectionMetrics: ReturnItemCollectionMetrics? = nil, returnValues: ReturnValue? = nil, tableName: String, updateExpression: String? = nil, updateItem: T) {
-            self.additionalAttributeNames = additionalAttributeNames
-            self.additionalAttributeValues = additionalAttributeValues
+        public init(conditionExpression: String? = nil, expressionAttributeNames: [String: String]? = nil, key: [String], returnConsumedCapacity: ReturnConsumedCapacity? = nil, returnItemCollectionMetrics: ReturnItemCollectionMetrics? = nil, returnValues: ReturnValue? = nil, tableName: String, updateExpression: String? = nil, updateItem: T) {
+            self.additionalAttributeNames = nil
+            self.additionalAttributeValues = nil
             self.conditionExpression = conditionExpression
             self.expressionAttributeNames = expressionAttributeNames
             self.key = key
@@ -360,6 +360,20 @@ extension DynamoDB {
             self.returnValues = returnValues
             self.tableName = tableName
             self.updateExpression = updateExpression
+            self.updateItem = updateItem
+        }
+
+        public init(additionalAttributeNames: [String: String]?, additionalAttributeValues: [String: AttributeValue]?, conditionExpression: String? = nil, key: [String], returnConsumedCapacity: ReturnConsumedCapacity? = nil, returnItemCollectionMetrics: ReturnItemCollectionMetrics? = nil, returnValues: ReturnValue? = nil, tableName: String, updateItem: T) {
+            self.additionalAttributeNames = additionalAttributeNames
+            self.additionalAttributeValues = additionalAttributeValues
+            self.conditionExpression = conditionExpression
+            self.expressionAttributeNames = nil
+            self.key = key
+            self.returnConsumedCapacity = returnConsumedCapacity
+            self.returnItemCollectionMetrics = returnItemCollectionMetrics
+            self.returnValues = returnValues
+            self.tableName = tableName
+            self.updateExpression = nil
             self.updateItem = updateItem
         }
 
--- a/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
+++ b/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
@@ -327,9 +327,9 @@ extension DynamoDB {
     }
 
     public struct UpdateItemCodableInput<T: Encodable & Sendable>: AWSEncodableShape {
-        /// In case expressionAttributeNames is nil, the content of additionalAttributeNames is merged with the generated expressionAttributeNames dictionary and will override existing values. This can be used to specify complex condtionExpression with not auto-generated attribute names.
+        /// In case expressionAttributeNames is nil, the content of additionalAttributeNames is merged with the generated expressionAttributeNames dictionary and will override existing values. This can be used to specify complex conditionExpression with not auto-generated attribute names.
         public let additionalAttributeNames: [String: String]?
-        /// The content of additionalAttributeValues is merged with expressionAttributeValues dictionary and will override existing values. This can be used to specify complex condtionExpression with not auto-generated attribute values.
+        /// The content of additionalAttributeValues is merged with expressionAttributeValues dictionary and will override existing values. This can be used to specify complex conditionExpression with not auto-generated attribute values.
         public let additionalAttributeValues: [String: AttributeValue]?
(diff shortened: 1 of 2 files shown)


adam-fowler:
Does this basically allow you to add conditions on fields that aren't included in the Codable type?

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
```diff
@@ -371,10 +377,20 @@ extension DynamoDB {
...
+            } else if let additionalAttributeNames {
+                let tmpAttributeNames: [String: String] = .init(item.keys.map { ("#\($0)", $0) }) { first, _ in return first }
+                expressionAttributeNames = tmpAttributeNames.merging(additionalAttributeNames, uniquingKeysWith: { _, new in new })
             } else {
                 expressionAttributeNames = .init(item.keys.map { ("#\($0)", $0) }) { first, _ in return first }
             }
-            let expressionAttributeValues: [String: AttributeValue] = .init(item.map { (":\($0.key)", $0.value) }) { first, _ in return first }
+
+            let expressionAttributeValues: [String: AttributeValue]
+            if let additionalAttributeValues {
```
adam-fowler, line 388: We still support swift 5.6, so this will need to be the old style `if let`

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
```diff
@@ -345,7 +349,9 @@ extension DynamoDB {
         /// Object containing item key and attributes to update
         public let updateItem: T
 
-        public init(conditionExpression: String? = nil, expressionAttributeNames: [String: String]? = nil, key: [String], returnConsumedCapacity: ReturnConsumedCapacity? = nil, returnItemCollectionMetrics: ReturnItemCollectionMetrics? = nil, returnValues: ReturnValue? = nil, tableName: String, updateExpression: String? = nil, updateItem: T) {
+        public init(additionalAttributeNames: [String: String]? = nil, additionalAttributeValues: [String: AttributeValue]? = nil, conditionExpression: String? = nil, expressionAttributeNames: [String: String]? = nil, key: [String], returnConsumedCapacity: ReturnConsumedCapacity? = nil, returnItemCollectionMetrics: ReturnItemCollectionMetrics? = nil, returnValues: ReturnValue? = nil, tableName: String, updateExpression: String? = nil, updateItem: T) {
```
adam-fowler, line 352: Can we create a separate `init` which include `additionalAttributeNames` and `additionalAttributeValues` without default values

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
```diff
@@ -371,10 +377,20 @@ extension DynamoDB {
             let expressionAttributeNames: [String: String]
             if let names = self.expressionAttributeNames, self.updateExpression != nil {
                 expressionAttributeNames = names
+            } else if let additionalAttributeNames {
+                let tmpAttributeNames: [String: String] = .init(item.keys.map { ("#\($0)", $0) }) { first, _ in return first }
+                expressionAttributeNames = tmpAttributeNames.merging(additionalAttributeNames, uniquingKeysWith: { _, new in new })
```
adam-fowler, line 382: So additional attribute names will override what is extracted from the Codable type? Is there a case where this is a valid situation?

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
```diff
@@ -371,10 +377,20 @@ extension DynamoDB {
             let expressionAttributeNames: [String: String]
             if let names = self.expressionAttributeNames, self.updateExpression != nil {
                 expressionAttributeNames = names
```
adam-fowler, line 379: Given you can already provide the `expressionAttributeNames` array is there a need for the `additionalAttributeNames` parameter.

Andrea-Scuderi:
> Does this basically allow you to add conditions on fields that aren't included in the Codable type?

Yes, correct.

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
Andrea-Scuderi: Not really, but it could be the only way to override the computed value. Do you think is better avoid overriding the Codable type?

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
Andrea-Scuderi: https://github.com/soto-project/soto/pull/671/commits/94c7cf4290d42d6b9c34b42c097e5dee23f5c200

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
Andrea-Scuderi: https://github.com/soto-project/soto/pull/671/commits/a1fdd48e476d4cc4fce14edf23ea0d86cbf46aa3

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
Andrea-Scuderi: https://github.com/soto-project/soto/pull/671/commits/a1fdd48e476d4cc4fce14edf23ea0d86cbf46aa3

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
adam-fowler: I was wondering what the intent was, maybe a comment should indicate whether the codable, or additional attribute takes precedence.

Andrea-Scuderi: I fixed the comment about additionalAttributeNames and additionalAttributeValues.

Andrea-Scuderi:
@adam-fowler Not sure why the PR fails on S3 tests though, they are not involved in the change.

adam-fowler:
> @adam-fowler Not sure why the PR fails on S3 tests though, they are not involved in the change.

localstack updated and broke something. I have a local fix, once that is in I'll tell you and you can merge main back into this branch

adam-fowler:
If you merge main in now it should work.

Andrea-Scuderi:
> If you merge main in now it should work.

I've merged could you give it a go?

Andrea-Scuderi:
> Hi Andrea, I have been looking at this again and was wondering the API seems an odd mix of Attributes and Codable types. Do you think you could replace the additional names and values with a Codable additionalAttributes. eg
> 
> ```swift
> init<AdditionalAttributes: Decodable>(additionAttributes: AdditionalAttributes, conditionExpression: ...) {
>     var item = try DynamoDBEncoder().encode(additionAttributes)
>     ...
> ```
> 
> I don't know if this makes sense, what do you think?

It looks promising, the problem of adding that in the init method is that it will require to add it to the UpdateItemCodableInput as well or I need to add a new struct.
```
public struct UpdateItemCodableInput<T: Encodable & Sendable, AdditionalAttributes: Decodable>: AWSEncodableShape
```

the alternative would be to add a just a new function
```
func createUpdateItemInput<AdditionalAttributes: Decodable>(additionAttributes: AdditionalAttributes) throws -> UpdateItemInput 
```

Which one you prefer?

Andrea-Scuderi:
> Hi Andrea, I have been looking at this again and was wondering the API seems an odd mix of Attributes and Codable types. Do you think you could replace the additional names and values with a Codable additionalAttributes. eg
> 
> ```swift
> init<AdditionalAttributes: Decodable>(additionAttributes: AdditionalAttributes, conditionExpression: ...) {
>     var item = try DynamoDBEncoder().encode(additionAttributes)
>     ...
> ```
> 
> I don't know if this makes sense, what do you think?

Nice one!
I fixed it in https://github.com/soto-project/soto/pull/671/commits/07fb3cd3937794044dfbf56fa7fb25ef952dfc7b

Could you run the CI?

adam-fowler (approved):
This looks good. Wish I had done the initialisers for all the Codable stuff in a similar manner, instead of adding new Codable request types. Might add them later and deprecate the current separate type with conversion function. What's your thoughts?

Andrea-Scuderi:
> This looks good. Wish I had done the initialisers for all the Codable stuff in a similar manner, instead of adding new Codable request types. Might add them later and deprecate the current separate type with conversion function. What's your thoughts?

Yes, it seems a good idea, but it's required to break compatibility.
I realised this solution after I tried to add the Encodable in the struct declaration.
I'd suggest to merge this and then iterate to improve that class.
Happy to help.

adam-fowler:
> I'd suggest to merge this and then iterate to improve that class.

Oh yes I agree there

PR #673, opened 2023-06-15
Fix additionalAttributeNames in UpdateItemCodableInput by mapping key array

Andrea-Scuderi:
#671 has introduced a new method to initialize UpdateItemCodableInput, but it has introduced an issue.

Consider the following code:
```swift 
private struct AdditionalAttributes: Encodable {
        let keyName: String
        let oldUpdatedAt: String
}
    
func updateItem<T: BreezeCodable>(item: T) async throws -> T {
        var item = item
        let oldUpdatedAt = item.updatedAt ?? ""
        let date = Date()
        item.updatedAt = date.iso8601
        let attributes = AdditionalAttributes(keyName: keyName, oldUpdatedAt: oldUpdatedAt)
        let input = try DynamoDB.UpdateItemCodableInput(
            additionalAttributes: attributes,
            conditionExpression: "attribute_exists(#keyName) AND #updatedAt = :oldUpdatedAt AND #createdAt = :createdAt",
            key: [keyName],
            tableName: tableName,
            updateItem: item
        )
        let _ = try await db.updateItem(input)
        return try await readItem(key: item.key)
}
```

The intention of the previous code is to update the item only when the key exists and when the record to update has the same createdAt and updatedAt data stored in DynamoDB. The update will change updatedAt with a new timestamp. In this way a client that doesn't have the same item cannot update the DynamoDB table and is forced to read the item before an update. (Optimistic locking)

But it leads to the following exception:

```
test_updateItem(): failed: caught error: "ValidationException: Value provided in ExpressionAttributeNames unused in expressions: keys: {#oldUpdatedAt}"
```

To fix the issue the code has be changed to include only the key in additionalAttributeNames and avoiding to add attributes that are not used.

The client code can be fixed by using:
```swift
private struct AdditionalAttributes: Encodable {
        let oldUpdatedAt: String
}
    
func updateItem<T: BreezeCodable>(item: T) async throws -> T {
        var item = item
        let oldUpdatedAt = item.updatedAt ?? ""
        let date = Date()
        item.updatedAt = date.iso8601
        let attributes = AdditionalAttributes(oldUpdatedAt: oldUpdatedAt)
        let input = try DynamoDB.UpdateItemCodableInput(
            additionalAttributes: attributes,
            conditionExpression: "attribute_exists(#\(keyName)) AND #updatedAt = :oldUpdatedAt AND #createdAt = :createdAt",
            key: [keyName],
            tableName: tableName,
            updateItem: item
        )
        let _ = try await db.updateItem(input)
        return try await readItem(key: item.key)
}
```

In the Unit Tests, the optimistic locking has been tested using `id` as key and `age` as condition to validate the record.

An alternative solution could be to add an explicit parameter to additionalAttributeNames.

### DIFF
--- a/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
+++ b/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
@@ -365,7 +365,7 @@ extension DynamoDB {
 
         public init<AdditionalAttributes: Encodable>(additionalAttributes: AdditionalAttributes, conditionExpression: String? = nil, key: [String], returnConsumedCapacity: ReturnConsumedCapacity? = nil, returnItemCollectionMetrics: ReturnItemCollectionMetrics? = nil, returnValues: ReturnValue? = nil, tableName: String, updateItem: T) throws {
             let attributes = try DynamoDBEncoder().encode(additionalAttributes)
-            self.additionalAttributeNames = .init(attributes.keys.map { ("#\($0)", $0) }) { first, _ in return first }
+            self.additionalAttributeNames = .init(key.map { ("#\($0)", $0) }) { first, _ in return first }
             self.additionalAttributeValues = .init(attributes.map { (":\($0.key)", $0.value) }) { first, _ in return first }
             self.conditionExpression = conditionExpression
             self.expressionAttributeNames = nil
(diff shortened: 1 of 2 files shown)


Andrea-Scuderi:
> Swift 5.8 failed https://github.com/soto-project/soto/actions/runs/5275592595/jobs/9541238015?pr=673#step:5:8784
> 
> Maybe localstack getting it wrong though

The age in the example was updated to 33 to prove that conditional update works when executed against the last update.
All seems ok.
Still in doubt if It's better to have explicit parameter for additionalAttributeNames.
For implementing optimistic locking this seems enough.

adam-fowler:
Just re-running the tests. I can't get them to fail locally with local stack.

PR #684, opened 2023-07-10
Strict concurrency

adam-fowler:


Package.swift
```diff
@@ -743,3 +749,11 @@ let package = Package(
         )
     ]
 )
+
+if ProcessInfo.processInfo.environment["STRICT_CONCURRENCY"] == "true" {
+    for target in package.targets {
+        if !target.isTest {
+            target.swiftSettings = [.unsafeFlags(["-Xfrontend", "-warn-concurrency"])]
```
0xTim, line 756: Shouldn't this be strict concurrency rather than warning
```suggestion
            target.swiftSettings = [.unsafeFlags(["-Xfrontend", "-strict-concurrency=complete"])]
```

adam-fowler: I didn't know about that one. Finding these flags is so bloody hard work, having them as errors makes life a lot easier

PR #686, opened 2023-07-19
Add support for decoding event streams

adam-fowler:
- Generated code for event stream parsing
- Removed custom S3 selectObjectContent as this is now support by the generic implementation in SotoCore

Related PRs https://github.com/soto-project/soto-core/pull/566, https://github.com/soto-project/soto-codegenerator/pull/74


Sources/Soto/Services/IoTEventsData/IoTEventsData_shapes.swift
```diff
@@ -1378,7 +1378,7 @@ extension IoTEventsData {
         }
 
         public func validate(name: String) throws {
-            try self.validate(self.timeInMillis, name: "timeInMillis", parent: name, max: 9223372036854775807)
+            try self.validate(self.timeInMillis, name: "timeInMillis", parent: name, max: -9223372036854775808)
```
0xTim, line 1381: Was this meant to switch to the minimum value for an Int64?

Sources/Soto/Services/KafkaConnect/KafkaConnect_shapes.swift
```diff
@@ -583,7 +583,7 @@ extension KafkaConnect {
         }
 
         public func validate(name: String) throws {
-            try self.validate(self.revision, name: "revision", parent: name, max: 9223372036854775807)
+            try self.validate(self.revision, name: "revision", parent: name, max: -9223372036854775808)
```
0xTim, line 586: As above

Sources/Soto/Services/KafkaConnect/KafkaConnect_shapes.swift
```diff
@@ -1630,7 +1630,7 @@ extension KafkaConnect {
...
```
0xTim, line 1633: As above

0xTim (changes requested):
Query around the flipping of max values for an Int64 - unsure if this was intentional or not but it doesn't feel like it

PR #691, opened 2023-09-02
S3 multipart upload of ByteBuffer

adam-fowler:
We had functions for loading an AsyncSequence of ByteBuffers and a file, but not a single ByteBuffer

### DIFF
--- a/Sources/Soto/Extensions/S3/S3+multipart.swift
+++ b/Sources/Soto/Extensions/S3/S3+multipart.swift
@@ -194,6 +194,41 @@ extension S3 {
         return downloaded
     }
 
+    /// Multipart upload of ByteBuffer to S3.
+    ///
+    /// Uploads buffer using multipart upload commands. If you want the function to not abort the multipart upload when it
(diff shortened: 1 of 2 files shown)


Sources/Soto/Extensions/S3/S3+multipart.swift
```diff
@@ -599,6 +633,37 @@ extension S3 {
...
+
+// Internal functions used by multipart upload
+extension S3 {
+    /// Do all the work for opening a file and closing it for MultiUpload function
+    func openFileForMultipartUpload(
+        filename: String,
+        logger: Logger,
+        on eventLoop: EventLoop,
+        threadPoolProvider: ThreadPoolProvider = .createNew,
+        uploadCallback: @escaping (NIOFileHandle, FileRegion, NonBlockingFileIO) async throws -> CompleteMultipartUploadOutput
```
0xTim, line 646: Question - if a closure is marked as `async` is it implicitly `@Sendable` because it's crossing a boundary?

0xTim: PS if not this probably wants to be `Sendable`

0xTim (approved):

PR #695, opened 2023-10-04
Use `NIOThreadPool.singleton` instead of `.createNew` in multipart upload

adam-fowler:


adam-fowler:
Looks like we have to do this on a background thread, as the first reference to `.singleton` will create it and if this isn't on a background thread we get a thread priority inversion warning.

0xTim (approved):

PR #696, opened 2023-10-04
7.0) Use `NIOThreadPool.singleton` instead of `.createNew` in multipart upload

adam-fowler:
7.x.x version of #695 . This one gets rid of `createNew` altogether


0xTim (approved):

adam-fowler:
I need to update this to reference the singleton on a background thread

PR #703, opened 2023-12-08
Replace runThrowingTask with syncAwait

adam-fowler:
runThrowingTask/runTask both required an EventLoop to run. Future changes will mean we will not have one available. 

As with runThrowingTask/runTask this function should only ever be used in tests as it blocks the current thread.

### DIFF
--- a/Tests/SotoTests/Services/ApiGatewayV2/APIGatewayV2Tests.swift
+++ b/Tests/SotoTests/Services/ApiGatewayV2/APIGatewayV2Tests.swift
@@ -47,22 +47,26 @@ class APIGatewayV2Tests: XCTestCase {
         }
         /// If we create a rest api for each test, when we delete them APIGateway will throttle
         /// and we will most likely not delete the all APIs so we create one API to be used by all tests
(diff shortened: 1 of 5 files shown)


Tests/SotoTests/Services/ApiGatewayV2/APIGatewayV2Tests.swift
```diff
@@ -47,22 +47,26 @@ class APIGatewayV2Tests: XCTestCase {
...
+                } catch {
+                    print("Failed to create APIGateway rest api, error: \(error)")
+                    throw error
+                }
             }
-        })
+        }.syncAwait()
     }
 
     override class func tearDown() {
```
0xTim, line 62: Why not use the async teardown and setup functions to avoid needing to putting it in a task and awaiting on it?

0xTim: Ah I missed that - yeah that's a massive API hole

0xTim (approved):

PR #710, opened 2024-03-29
Add S3.generatePresignedPost for HTML form based uploads

nicksloan:
Adds a generatePresignedPost method to the S3 service, similar to [what is available in Boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/generate_presigned_post.html).

Amazon docs for this type of request live [here](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-UsingHTTPPOST.html).

Remaining work:
- [x] Add error handling
- [x] Add documentation
- [x] Add tests

### DIFF
--- a/Sources/Soto/Extensions/S3/S3+presignedPost.swift
+++ b/Sources/Soto/Extensions/S3/S3+presignedPost.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Soto for AWS open source project
 //
-// Copyright (c) 2017-2021 the Soto project authors
+// Copyright (c) 2024 the Soto project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -166,7 +166,7 @@ extension S3 {
         let shortDate = self.shortDateFormat(date: date)
 
         let clientCredentials = try await client.getCredential()
-        let presignedPostCredential = try await getPresignedPostCredential(date: shortDate, accessKeyId: clientCredentials.accessKeyId)
+        let presignedPostCredential = getPresignedPostCredential(date: shortDate, accessKeyId: clientCredentials.accessKeyId)
 
         var keyCondition: PostPolicyCondition
         let suffix = "${filename}"
@@ -196,8 +196,8 @@ extension S3 {
         fields["Policy"] = stringToSign
 
         // Create the signature and add to fields
-        let signingKey = try await signingKey(date: shortDate, secretAccessKey: clientCredentials.secretAccessKey)
-        let signature = try await getSignature(policy: stringToSign, signingKey: signingKey)
+        let signingKey = signingKey(date: shortDate, secretAccessKey: clientCredentials.secretAccessKey)
+        let signature = getSignature(policy: stringToSign, signingKey: signingKey)
         fields["x-amz-signature"] = signature
 
         // Create the response
@@ -206,24 +206,23 @@ extension S3 {
         return presignedPostResponse
     }
 
-    func signingKey(date: String, secretAccessKey: String) async throws -> SymmetricKey {
-        let credentials = try await client.getCredential()
+    func signingKey(date: String, secretAccessKey: String) -> SymmetricKey {
         let name = config.signingName
         let region = config.region.rawValue
 
-        let kDate = HMAC<SHA256>.authenticationCode(for: [UInt8](date.utf8), using: SymmetricKey(data: Array("AWS4\(credentials.secretAccessKey)".utf8)))
+        let kDate = HMAC<SHA256>.authenticationCode(for: [UInt8](date.utf8), using: SymmetricKey(data: Array("AWS4\(secretAccessKey)".utf8)))
         let kRegion = HMAC<SHA256>.authenticationCode(for: [UInt8](region.utf8), using: SymmetricKey(data: kDate))
         let kService = HMAC<SHA256>.authenticationCode(for: [UInt8](name.utf8), using: SymmetricKey(data: kRegion))
         let kSigning = HMAC<SHA256>.authenticationCode(for: [UInt8]("aws4_request".utf8), using: SymmetricKey(data: kService))
         return SymmetricKey(data: kSigning)
     }
 
-    func getSignature(policy: String, signingKey key: SymmetricKey) async throws -> String {
+    func getSignature(policy: String, signingKey key: SymmetricKey) -> String {
         let signature = HMAC<SHA256>.authenticationCode(for: [UInt8](policy.utf8), using: key).hexDigest()
         return signature
     }
 
-    func getPresignedPostCredential(date: String, accessKeyId: String) async throws -> String {
+    func getPresignedPostCredential(date: String, accessKeyId: String) -> String {
         let region = config.region.rawValue
         let service = config.signingName
 
--- a/Sources/Soto/Extensions/S3/S3+presignedPost.swift
+++ b/Sources/Soto/Extensions/S3/S3+presignedPost.swift
@@ -12,8 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-import Logging
 import Foundation
+import Logging
 
 import Crypto
 @_spi(SotoInternal) import SotoSignerV4
@@ -48,15 +48,15 @@ extension S3 {
 
         public func encode(to encoder: Encoder) throws {
             switch self {
-                case let .match(field, value):
-                    let condition = [field: value]
-                    var container = encoder.singleValueContainer()
-                    try container.encode(condition)
+            case .match(let field, let value):
+                let condition = [field: value]
+                var container = encoder.singleValueContainer()
+                try container.encode(condition)
 
-                case let .rule(rule, field, value):
-                    let condition = [rule, field, value]
-                    var container = encoder.singleValueContainer()
-                    try container.encode(condition)
+            case .rule(let rule, let field, let value):
+                let condition = [rule, field, value]
+                var container = encoder.singleValueContainer()
+                try container.encode(condition)
             }
         }
     }
@@ -67,15 +67,15 @@ extension S3 {
         let fields: [String: String]
     }
 
-    ///  Builds the url and the form fields used for a presigned s3 post 
+    ///  Builds the url and the form fields used for a presigned s3 post
     /// - Parameters:
-    ///   - key: Key name, optionally add ${filename} to the end to attach the submitted filename. Note that key related conditions and fields are filled out for you and should not be included in the Fields or Conditions parameter. 
+    ///   - key: Key name, optionally add ${filename} to the end to attach the submitted filename. Note that key related conditions and fields are filled out for you and should not be included in the Fields or Conditions parameter.
     ///   - bucket: The name of the bucket to presign the post to. Note that bucket related conditions should not be included in the conditions parameter.
     ///   - fields: A dictionary of prefilled form fields to build on top of. Elements that may be included are acl, Cache-Control, Content-Type, Content-Disposition, Content-Encoding, Expires, success_action_redirect, redirect, success_action_status, and x-amz-meta-.
-    /// 
+    ///
     ///     Note that if a particular element is included in the fields dictionary it will not be automatically added to the conditions list. You must specify a condition for the element as well.
     ///   - conditions: A list of conditions to include in the policy. Each element can be either a match or a rule. For example:
-    /// 
+    ///
     ///     ```
     ///     [
     ///         .match("acl", "public-read"), .rule("content-length-range", "2", "5"), .rule("starts-with", "$success_action_redirect", "")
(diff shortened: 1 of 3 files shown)


Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,194 @@
...
+
+    /// An encodable struct that represents the URL and form fields to use in a presigned POST request to S3
+    public struct PresignedPostResponse: Encodable {
+        let url: URL
+        let fields: [String: String]
+    }
+
+    ///  Builds the url and the form fields used for a presigned s3 post 
+    /// - Parameters:
+    ///   - key: Key name, optionally add ${filename} to the end to attach the submitted filename. Note that key related conditions and fields are filled out for you and should not be included in the Fields or Conditions parameter.
```
adam-fowler, line 59: Comment lines are too long. Can we add new lines where applicable

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,194 @@
...
+    ///         .match("acl", "public-read"), .rule("content-length-range", "2", "5"), .rule("starts-with", "$success_action_redirect", "")
+    ///     ]
+    ///     ```
+    ///
+    ///     Conditions that are included may pertain to acl, content-length-range, Cache-Control, Content-Type, Content-Disposition, Content-Encoding, Expires, success_action_redirect, redirect, success_action_status, and/or x-amz-meta-.
+    ///
+    ///     Note that if you include a condition, you must specify the a valid value in the fields dictionary as well. A value will not be added automatically to the fields dictionary based on the conditions.
+    ///   - expiresIn: The number of seconds the presigned post is valid for.
+    /// - Returns: An encodable PresignedPostResponse with two properties: url and fields. Url is the url to post to. Fields is a dictionary filled with the form fields and respective values to use when submitting the post.
+    public func generatePresignedPost(key: String, bucket: String, fields: [String: String] = [:], conditions: [PostPolicyCondition] = [], expiresIn: TimeInterval, now: Date = Date.now) async throws -> PresignedPostResponse {
```
adam-fowler, line 77: Can we change `now` to `date`.

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,194 @@
...
+        fields["x-amz-signature"] = signature
+
+        // Create the response
+        let presignedPostResponse = PresignedPostResponse(url: url, fields: fields)
+
+        return presignedPostResponse
+    }
+
+    private func signingKey(date: String) async throws -> SymmetricKey {
+        let credentials = try await client.getCredential()
```
adam-fowler, line 148: It looks like you call `client.getCredential` twice. once here and once in `getCredential`. This can be an expensive function, you should probably reuse the value

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,194 @@
...
+        return SymmetricKey(data: kSigning)
+    }
+
+    private func getSignature(policy: String, date: String) async throws -> String {
+        let key = try await signingKey(date: date)
+        let signature = HMAC<SHA256>.authenticationCode(for: [UInt8](policy.utf8), using: key).hexDigest()
+        return signature
+    }
+
+    private func getCredential(date: String) async throws -> String {
```
adam-fowler, line 165: This is a slightly confusing name. `getCredential` that calls `getCredential`

Tests/SotoTests/Services/S3/S3ExtensionTests.swift
```diff
@@ -370,4 +370,56 @@ extension S3Tests {
...
+        let expectedDate = "20151229T000000Z"
+
+        XCTAssertEqual(presignedPost.url, URL(string: expectedURL))
+        XCTAssertEqual(presignedPost.fields["Policy"], expectedPolicy)
+        XCTAssertEqual(presignedPost.fields["x-amz-signature"], expectedSignature)
+        XCTAssertEqual(presignedPost.fields["x-amz-credential"], expectedCredential)
+        XCTAssertEqual(presignedPost.fields["x-amz-algorithm"], expectedAlgorithm)
+        XCTAssertEqual(presignedPost.fields["x-amz-date"], expectedDate)
+
+        try? await client.shutdown()
```
adam-fowler, line 422: Move this into a `defer` block below where you create the client. Otherwise if `s3.generatePresignedPost` throws an error the error you will see is the AWSClient not closing down correctly which can be confusing.

adam-fowler (commented):
Here are some initial comments. I haven't actually tested this code yet

nicksloan:
@adam-fowler Thanks for the feedback. The test definitely won’t pass. The inputs from the AWS docs don’t really work for a reproducible output. Working on a better test case.

adam-fowler:
You'll need to rebase 7.x.x to get this compiling again by the way

nicksloan:
> You'll need to rebase 7.x.x to get this compiling again by the way

Rebased this branch, and now I'm getting a lot of errors in STS.

nicksloan:
I have to approach testing this differently, but I have a plan. Update coming sometime tomorrow afternoon.

adam-fowler:
Why can't you use the example to verify you are calculating the correct signature?

nicksloan:
The example string to sign in their docs is a base 64 encoded JSON object, but the JSON object they used has arbitrary whitespace in it, so it would be absurd to try to make this code produce the exact same string.

I can make use of the examples from the docs to test at a more granular level. If I test the getSignature method in isolation (rather than generatePresignedPost as a whole), I can supply the string to sign directly, rather than generating it. I’ll push an updated test that does  that once I’m at my desk today.

The other thing I tried is using Boto3 to generate known to work values from example credentials, but Python preserves order in dictionaries, which means the properties of the JSON object end up in a different order. Same problem, we’d have to make our code worse to match the output exactly, for absolutely no gain outside of the test suite.

In addition to testing getSignature, I’ll test getPresignedPostCredential, and I’ll have to test that generatePresignedPost produces the right URL, and a fields dictionary and post policy that have all the right keys. If I test all of that, I’m confident that the code is well covered. I’m about halfway done with these tests, and should be able to finish this morning. Thanks for your patience.

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,251 @@
...
+        let credentials = try await client.getCredential()
+        let name = config.signingName
+        let region = config.region.rawValue
+
+        let kDate = HMAC<SHA256>.authenticationCode(for: [UInt8](date.utf8), using: SymmetricKey(data: Array("AWS4\(credentials.secretAccessKey)".utf8)))
+        let kRegion = HMAC<SHA256>.authenticationCode(for: [UInt8](region.utf8), using: SymmetricKey(data: kDate))
+        let kService = HMAC<SHA256>.authenticationCode(for: [UInt8](name.utf8), using: SymmetricKey(data: kRegion))
+        let kSigning = HMAC<SHA256>.authenticationCode(for: [UInt8]("aws4_request".utf8), using: SymmetricKey(data: kService))
+        return SymmetricKey(data: kSigning)
+    }
```
nicksloan, line 219: This function is largely lifted from soto-core, but that version is not quite reusable as-is here. Felt like a bigger lift than a first time contributor should take on to rearrange things there.

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,251 @@
...
+        return signature
+    }
+
+    func getPresignedPostCredential(date: String, accessKeyId: String) async throws -> String {
+        let region = config.region.rawValue
+        let service = config.signingName
+
+        let credential = "\(accessKeyId)/\(date)/\(region)/\(service)/aws4_request"
+        return credential
+    }
```
nicksloan, line 232: I think this also exists in some form in soto-core. See previous comment.

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,251 @@
...
+
+    private func longDateFormat(date: Date) -> String {
+        let dateFormatter = DateFormatter()
+        dateFormatter.locale = Locale(identifier: "en_US_POSIX")
+        dateFormatter.dateFormat = "yyyyMMdd'T'HHmmss'Z'"
+        dateFormatter.timeZone = TimeZone(secondsFromGMT: 0)
+
+        return dateFormatter.string(from: date)
+    }
```
nicksloan, line 250: I bet you're producing these date formats elsewhere, but I couldn't find it. If there is a way to consolidate, that would be great.

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,251 @@
...
+        conditions: [PostPolicyCondition] = [],
+        expiresIn: TimeInterval
+    ) async throws -> PresignedPostResponse {
+        try await self.generatePresignedPost(
+            key: key,
+            bucket: bucket,
+            fields: fields,
+            conditions: conditions,
+            expiresIn: expiresIn,
+            date: Date.now
```
adam-fowler, line 126: `Date.now` is not available in macOS 10.15 
```suggestion
            date: Date()
```

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    // Private API adds date argument for testing
+    func generatePresignedPost(
+        key: String,
+        bucket: String,
+        fields: [String: String] = [:],
+        conditions: [PostPolicyCondition] = [],
+        expiresIn: TimeInterval,
+        date: Date = Date.now
```
adam-fowler, line 137: `Date.now` is not available in macOS 10.15 
```suggestion
        date: Date = Date()
```

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,251 @@
...
+        let kSigning = HMAC<SHA256>.authenticationCode(for: [UInt8]("aws4_request".utf8), using: SymmetricKey(data: kService))
+        return SymmetricKey(data: kSigning)
+    }
+
+    func getSignature(policy: String, signingKey key: SymmetricKey) async throws -> String {
+        let signature = HMAC<SHA256>.authenticationCode(for: [UInt8](policy.utf8), using: key).hexDigest()
+        return signature
+    }
+
+    func getPresignedPostCredential(date: String, accessKeyId: String) async throws -> String {
```
adam-fowler, line 226: You'll need to fix the call sites as well
```suggestion
    func getPresignedPostCredential(date: String, accessKeyId: String) -> String {
```

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,251 @@
...
+        let region = config.region.rawValue
+
+        let kDate = HMAC<SHA256>.authenticationCode(for: [UInt8](date.utf8), using: SymmetricKey(data: Array("AWS4\(credentials.secretAccessKey)".utf8)))
+        let kRegion = HMAC<SHA256>.authenticationCode(for: [UInt8](region.utf8), using: SymmetricKey(data: kDate))
+        let kService = HMAC<SHA256>.authenticationCode(for: [UInt8](name.utf8), using: SymmetricKey(data: kRegion))
+        let kSigning = HMAC<SHA256>.authenticationCode(for: [UInt8]("aws4_request".utf8), using: SymmetricKey(data: kService))
+        return SymmetricKey(data: kSigning)
+    }
+
+    func getSignature(policy: String, signingKey key: SymmetricKey) async throws -> String {
```
adam-fowler, line 221: You'll need to fix the call sites as well
```suggestion
    func getSignature(policy: String, signingKey key: SymmetricKey) -> String {
```

Sources/Soto/Extensions/S3/S3+presignedPost.swift
adam-fowler: This is fine as is

Sources/Soto/Extensions/S3/S3+presignedPost.swift
adam-fowler: I'm going to unresolved this until the fixed code is submitted

Sources/Soto/Extensions/S3/S3+presignedPost.swift
adam-fowler: Have you thought of using the `ISO8601DateFormatter`

Sources/Soto/Extensions/S3/S3+presignedPost.swift
```diff
@@ -0,0 +1,251 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Soto for AWS open source project
+//
+// Copyright (c) 2017-2021 the Soto project authors
```
adam-fowler, line 5:
```suggestion
// Copyright (c) 2024 the Soto project authors
```

nicksloan:
@adam-fowler Ready for another review.

adam-fowler (approved):
Looks good
Have you tested with real assets?

nicksloan:
> Looks good Have you tested with real assets?

I did test with real assets and credentials, but not from outside of the package until today. Submitted a small follow-up PR and have confirmed that everything works as intended once that is merged.

PR #716, opened 2024-04-23
fix: handle session tokens in generatePresignedPost

nicksloan:
Checks for the presence of a session token, and adds it to the presigned-POST request if it exists.


adam-fowler:
@nicksloan Have you tested this in the real world yet?

nicksloan:
@adam-fowler Just confirmed it works.

adam-fowler (approved):
looks good

Issue #730, opened 2024-09-17
Swift 6 warnings


MahdiBM:
**Describe the bug**
When building on `swift:6.0-noble` image, I get these warnings:
```
warning: 'soto': found 774 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
/__w/my-app/my-app/.build/checkouts/soto/Sources/Soto/Services/RDSData/RDSData_api.swift
/__w/my-app/my-app/.build/checkouts/soto/Sources/Soto/Extensions/CognitoIdentity/CognitoIdentity+CredentialProvider.swift
/__w/my-app/my-app/.build/checkouts/soto/Sources/Soto/Services/Kinesis/Kinesis_shapes.swift
/__w/my-app/my-app/.build/checkouts/soto/Sources/Soto/Services/ElasticsearchService/ElasticsearchService_api.swift
```

**To Reproduce**
Build on `swift:6.0-noble`

**Expected behavior**
No warnings

**Setup (please complete the following information):**
 - OS: Linux
 - Version of soto: latest

**Additional context**

adam-fowler:
This is not something specific to Swift 6. It's been around for years. It's a bug in SwiftPM that seems to be at the bottom of the pile. 

Basically SwiftPM doesn't like the structure of Soto where a library can be spread across two folders and have other libraries folder structure interleaved with it. I specifically say in the manifest only look in these folders, but it still complains about files outside of them.

MahdiBM:
Hmm perhaps Swift 6 is emitting the warning more prominently then.
I can believe that it was already there even before Swift 6, and I just didn't notice it too much.

adam-fowler:
It looks like this wasn't being reported for users of soto in v5.10 and earlier. It only appeared for developers of the package. This has changed in v6. I have mentioned this in the bug and hopefully we'll see some resolution. If not I'll be forced into merging #732 and creating another major release

MahdiBM:
This doesn't seem to stop there, and I think it's a general issue with warnings propagation of dependency Package.swifts.
I also see a bunch of these from other dependencies.
```
[...]

PR #732, opened 2024-09-21
Add extension targets

adam-fowler:
The issue with SwiftPM complaining about files that are in the target `path` but not in the `sources` folders is not going away. I posted this bug on feedback assistant 3 or 4 years back and then in the SwiftPM repo last year https://github.com/swiftlang/swift-package-manager/issues/6440 and there doesn't seem to be any movement on it. In actual fact now it has got worse as it is being reporting for users of soto and not just those compiling soto. See #730 

This PR would get around this by add new extension targets `SotoS3Extension` `SotoSTSExtension` etc but it is a breaking change and would require a soto v8.


0xTim (approved):

MahdiBM:
I'm curious to know what is the breaking change 🙂

adam-fowler:
To use the multipart extension code you would need to import SotoS3Extension.

MahdiBM:
So ... a @_exported wouldn't salvage the situation?

adam-fowler:
That'd produce a weird dynamic where SotoS3Extension is dependent on SotoS3 but is imported into SotoS3Extension, so SotoS3 becomes dependent on SotoS3Extension. Not sure that is possible. 

I guess you could have a SotoS3Service target that holds the generated code, SotoS3Extension that holds the extension code and then SotoS3 that imports both with `@_exported`. That might work but it's bloody horrible.

MahdiBM:
Honestly I don't see how that's horrible 😅
Makes a fine sense to me, specially to avoid a new major version.

I do see the difference but generally speaking, "umbrella modules" aren't rare. NIO for example.

Not that I'll be bothered with a new major version, but it feels awkward to have a new major version just for this, so I thought I take my shot to see if we can avoid that.

It's fine either way, for me.

PR #738, opened 2024-10-16
Export generated service files from extensions target

adam-fowler:
For services with  extension code, 
- rename the target with the generated code to `_Soto{{ServiceName}}Generated`.
- Add new target with correct name `Soto{{ServiceName}}` and add extension code to it.
- Add `@_exported import _Soto{{ServiceName}}Generated` in new target

This came out of discussions in #732


Sources/Soto/Extensions/DynamoDB/exports.swift
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Soto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@_exported import _SotoDynamoDBGenerated
```
MahdiBM, line 15: Might need some changes for proper documentation propagation. I don't have too much experience with that so I'll defer to @0xTim who has had to deal with a bunch of exported modules in Vapor docs.

adam-fowler: We don't use docc on the service files. So not an issue.

PR #742, opened 2024-11-22
adds support for compiling against Musl

t089:


adam-fowler:
I fixed this for Soto-core, missed I needed it here as well

adam-fowler:
Thanks for that

PR #756, opened 2025-03-07
DynamoDB Codable date encoding strategy

adam-fowler:
- Add encoding/decoding strategy to DynamoDBEncoder/Decoder
- Add encoder/decoder parameter to DynamoDB codable calls
- BREAKING CHANGE: Move encoding of additionalAttributes in `UpdateItemCodableInput` to `updateItem` call instead of `init`. This means `UpdateItemCodableInput` has an additional generic parameter

### DIFF
--- a/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
+++ b/Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
@@ -26,10 +26,11 @@ extension DynamoDB {
     /// Invalid Requests with empty values will be rejected with a `ValidationException` exception.</p> <note> <p>To prevent a new item from replacing an existing item, use a conditional expression that contains the `attribute_not_exists` function with the name of the attribute being used as the partition key for the table. Since every record must contain that attribute, the `attribute_not_exists` function will only succeed if no matching item exists.</p> </note> <p>For more information about `PutItem`, see <a href="https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html">Working with Items</a> in the <i>Amazon DynamoDB Developer Guide</i>.
     public func putItem(
         _ input: PutItemCodableInput<some Encodable>,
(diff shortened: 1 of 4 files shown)


nicksloan:
Haven't had a chance to run it just yet, but looking things over it seems like it'll work. I'll try to run it today or tomorrow.

It's a shame that you have to specify the encoder and decoder with the endpoint calls, but it'd be a much bigger change otherwise. In practice, we'll probably want to wrap the DynamoDB service in our code so that we can specify the modified encoder/decoder project wide in one place.

nicksloan:
@adam-fowler This is great! Got a chance to run it, and it works. Thank you!

Sources/Soto/Extensions/DynamoDB/DynamoDB+Codable.swift
```diff
@@ -221,11 +228,9 @@ extension DynamoDB {
         }
     }
 
-    public struct UpdateItemCodableInput<T: Encodable & Sendable>: AWSEncodableShape {
-        /// In case expressionAttributeNames is nil, the content of additionalAttributeNames is merged with the generated expressionAttributeNames dictionary and will override existing values. This can be used to specify complex conditionExpression with not auto-generated attribute names.
-        public let additionalAttributeNames: [String: String]?
+    public struct UpdateItemCodableInput<T: Encodable & Sendable, Additional: Encodable & Sendable>: AWSEncodableShape {
         /// The content of additionalAttributeValues is merged with expressionAttributeValues dictionary and will override existing values. This can be used to specify complex conditionExpression with not auto-generated attribute values.
-        public let additionalAttributeValues: [String: AttributeValue]?
```
0xTim, line 228: This is an API break isn't it?

adam-fowler: Yes it is. I can't see a better way to do this without crippling the API. Given the way people use the Request structures I thought I could get away. I will obviously report the break in the release notes

0xTim (approved):

PR #763, opened 2025-04-15
Add error code to error type map

adam-fowler:
This PR goes hand in hand with https://github.com/soto-project/soto-core/pull/649

Makes extended error information available via `AWSErrorType.context.extendedError` if available eg
```swift
do
    let updateRequest = try DynamoDB.UpdateItemCodableInput(...)
    _ = try await Self.dynamoDB.updateItem(updateRequest, logger: TestEnvironment.logger)
} catch let error as DynamoDBErrorType where error == .conditionalCheckFailedException {
    if let conditionError = error.context?.extendedError as? DynamoDB.ConditionalCheckFailedException {
        ...
    }
}
```


Package.swift
```diff
@@ -433,7 +433,7 @@ let package = Package(
         .library(name: "SotoXRay", targets: ["SotoXRay"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/soto-project/soto-core.git", from: "7.4.0")
+        .package(url: "https://github.com/soto-project/soto-core.git", branch: "custom-errors")
```
adam-fowler, line 436: Remember to set version before merging

0xTim (approved):

PR #767, opened 2025-05-27
Android support

gevartosky:
Goals ⚽

The goal of this pull request is to add Android support to soto.

The following changes were verified on the swift-6.1 toolchain built for
Android.

Implementation Details 🚧

core should work flawlessly against an android swift toolchain
(tested with version 6.1) with the same limitations and context as the
current support on Linux + Windows.

Testing Details 🔍

No new unit tests were added.

tested on android api level 30

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
   android:
     name: Android
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 35
     strategy:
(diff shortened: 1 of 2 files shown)


adam-fowler:
I'd rather not add an additional CI target for Soto. Each one takes 45 minutes.

gevartosky:
Makes sense, I noticed its taking 25+ to build.
Ok I'll remove the action and keep the `os(Android)` import

adam-fowler (approved):

